PR #5, opened 2016-03-06
load_template now accepts **args as it feels more natural

dbarrosop:
@mirceaulinic can you review this? What do you think?

The idea is to be able to treat this method in a more natural way, passing what I want in the template as named arguments:

```
driver.load_template('set_ntp_peers', peers=['1.1.1.1', '2.2.2.2'])
```

Alternatively, I wouldn't mind passing a dict, something like:

```
driver.load_template('set_ntp_peers', {'peers': ['1.1.1.1', '2.2.2.2']})
```

So in the template you can do:

```
    {% for peer in peers %}
      peer {{peer}}
    {% endfor %}
```

Instead of:

```
    {% for peer in template_vars %}
      peer {{peer}}
    {% endfor %}
```

Thoughts?

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -135,7 +135,7 @@ class NetworkDriver:
         """
         raise NotImplementedError
 
-    def load_template(self, template_name, template_vars):
+    def load_template(self, template_name, **template_vars):
         """
         Will load a templated configuration on the device.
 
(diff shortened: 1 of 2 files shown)


mirceaulinic:
Nice! Definitely a big improvement :+1:

PR #7, opened 2016-03-10
replaces static loader with dynamic loader

privateip:
This commit replaces the static loading of drivers with a dynamic loader.  The
dynamic loader can load modules using either a full path namespace
or a single module name.  If the provided name is not a full path
namespace, then the module name is prepended with the default namespace.

For example, calling get_network_driver('foo') will be translated to
'napalm_foo.foo' but get_network_driver('napalm.foo') will not be
modified.

### DIFF
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -14,33 +14,145 @@
 
 """napalm_base package."""
 
-
-def _get_eos_driver():
-    from napalm_eos.eos import EOSDriver
-    return EOSDriver
+import sys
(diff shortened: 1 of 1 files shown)


dbarrosop:
I haven't fully reviewed the PR yet but I have two initial comments:

1. Testing. I guess we will need to include a fake module or something to actually test this code added.
2. Wouldn't be better to have the `load_driver ` method in `__init__.py` ? It's consistent with the way base is also structured : )

PR #8, opened 2016-03-15
Moved load_template() and textfsm_extractor() in helpers module

mirceaulinic:
Also need to move the templates in all napalm_*:
Noticed they still reside under /templates/[DRIVER_NAME]/ instead of /templates/

### DIFF
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -83,7 +83,6 @@ def load_module(name):
     :raises ImportError:
         If the module could not be loaded
     """
-
     try:
         mod = None
         mod = sys.modules[name]
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -83,6 +83,7 @@ def load_module(name):
     :raises ImportError:
         If the module could not be loaded
     """
+
     try:
         mod = None
         mod = sys.modules[name]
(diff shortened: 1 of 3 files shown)


dbarrosop:
Some of the stuff here is obsoleted by the work here: https://github.com/napalm-automation/napalm-base/pull/11

dbarrosop:
In addition, this would allow us to remove gtextfsm and jinja2 from the requirements file and just add them where needed, right?

mirceaulinic:
Yes, but it would not be entirely correct. We define in napalm_base functions that have their dependencies. In napalm_* we use those functions without the having to know their content.

dbarrosop:
I agree but I would prefer not having to install requirements that I don't need. Anyway, in this case I think it's not a big deal, jinja2 and gtextfsm are used by most drivers (or should at least).

Can you rebase `napalm_base/__init__.py`?

dbarrosop:
> Noticed they still reside under /templates/[DRIVER_NAME]/ instead of /templates/

Really? I was convinced I moved them. I checked a few drivers and I can see they are located under ``napalm_driver/templates/``. Which one is missing?

mirceaulinic:
I think only junos had some issues, but looks OK now. I will rebase https://github.com/napalm-automation/napalm-junos/pull/8 too

mirceaulinic:
I'm afraid I do not understand: ```The command "nosetests ./test/unit/TestGetNetworkDriver.py" exited with 1.```.
Did the function ```load_driver``` exist before and I was not aware?

mirceaulinic:
Just saw #11. I let you fix that issue and afterwards I will rebase.

PR #9, opened 2016-03-15
Define get_probes_config() and get_probes_results()

mirceaulinic:
This refers to RPM probes on JunOS and SLA probes on IOS-XR.
On EOS is might be possible to add custom probes. Will investigate that.

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -946,16 +946,132 @@ class NetworkDriver:
             * acl (string) # acl number or name
             * mode (string) # read-write (rw), read-only (ro)
 
-        Example Output:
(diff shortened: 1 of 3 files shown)


mirceaulinic:
Here it will be slightly painful when waiting for the models and for the tests be part of the base and only afterwards to propose the body of the method.
But at least, like that we can have better control and force the user to propose the header of the method before anything else.

mirceaulinic:
Implementation in: https://github.com/napalm-automation/napalm-iosxr/pull/1 and https://github.com/napalm-automation/napalm-junos/pull/6

PR #10, opened 2016-03-16
Add ping() and traceroute()

mirceaulinic:


napalm_base/base.py
```diff
@@ -1079,3 +1079,124 @@ def get_probes_results(self):
...
+
+        :param destination: Host or IP Address of the destination
+        :param source (optional): Use a specific IP Address to execute the traceroute
+        :param ttl (optional): Maimum number of hops
+        :param timeout (optional): Number of seconds to wait for response
+
+        Output dictionary has one of the following keys:
+
+            * success
+            * error
```
dbarrosop: How do you define success/error? A CLI command issue? Not being able to reach destination?

napalm_base/base.py
```diff
@@ -1079,3 +1079,124 @@ def get_probes_results(self):
...
+
+            * rtt (float)
+            * ip_address (str)
+            * host_name (str)
+
+        Example output::
+
+            {
+                'success': {
+                    1: {
```
dbarrosop: Shouldn't this be a list of dicts? How do you represent something like hops 6 and 7?

```
traceroute to 8.8.8.8 (8.8.8.8), 64 hops max, 52 byte packets
 1  192.168.1.1 (192.168.1.1)  1.870 ms  2.010 ms  1.835 ms
 2  ssn-rt3-11-46-39-103-1.stosn.net (46.39.103.1)  2.625 ms  1.959 ms  1.927 ms
 3  ssn-rt2-01-42-10g.stosn.net (178.251.129.18)  2.063 ms  2.307 ms  2.344 ms
 4  ssn-rt1-01-31-20g.stosn.net (178.251.129.13)  2.435 ms  3.501 ms  16.990 ms
 5  as15169-te-tc1.sthix.net (192.121.80.47)  2.653 ms  2.774 ms  2.953 ms
 6  216.239.40.27 (216.239.40.27)  6.370 ms
    216.239.54.181 (216.239.54.181)  3.362 ms  3.376 ms
 7  72.14.234.89 (72.14.234.89)  3.487 ms
    209.85.242.47 (209.85.242.47)  4.878 ms
    74.125.37.129 (74.125.37.129)  2.945 ms
 8  google-public-dns-a.google.com (8.8.8.8)  4.099 ms  3.122 ms  3.620 ms
```

napalm_base/base.py
mirceaulinic: Error is when the device is unable to execute the traceroute, most often because either not able to reach destination, either unable to bind the source address, but not limited to.

napalm_base/base.py
mirceaulinic: For 6, I would consider only 216.239.40.27 and for 7 only 72.14.234.89. On IOS-XR, I force to one single probe per hop.
Do you think it would be more helpful to keep the default value of three probes/hop and return a list?
Anyway, on IOS-XR the XML output is more rather a text output where they threw some random XML tags: https://github.com/mirceaulinic/napalm-iosxr/blob/CF-TRACEROUTE/test/unit/iosxr/mock_data/_Set__Action__TraceRoute__IPV4__Destination_8.8.8.8__Destination__Probe_1__Probe___IPV4___TraceRoute___Action___Set_.rpc

mirceaulinic:
Closing, will revisit this later.

PR #11, opened 2016-03-21
get_network_driver will attempt to find the driver if load_driver doesnt exist

dbarrosop:
@privateip could you take a look? I made a few changes, the idea is as follow:

1. First I made most methods `internal`.
2. I changed how `get_network_driver` works. Now it will:
   1. If `load_driver` exists in `module` it returns whatever that method returns.
   2. If `load_driver` doesn't exist it will inspect all objects in `module` and return the first class that is a subclass of `NetworkDriver`.

This way you can either add `load_driver` to your module if you need/want to do something when fetching the driver or you can just add the line `from eos import EOSDriver`.

### DIFF
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -124,7 +124,7 @@ def _napalm_namespace(module):
     """
     if '.' in module:
         return module
-    return 'napalm_{}'.format(module, module)
+    return 'napalm_{}'.format(module)
 
 
 def _discover_driver_from_module(module):
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -16,10 +16,13 @@
 
 import sys
 import imp
+import inspect
+
+from base import NetworkDriver
 
 
-def import_module(name):
-    """Imports a module into the current runtime environment
+def _import_module(name):
+    """Import a module into the current runtime environment.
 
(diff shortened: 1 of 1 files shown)


napalm_base/__init__.py
```diff
@@ -121,11 +124,22 @@ def napalm_namespace(module):
     """
     if '.' in module:
         return module
-    return 'napalm_{}.{}'.format(module, module)
+    return 'napalm_{}'.format(module, module)
```
privateip: Did you mean to drop the second placeholder?

privateip:
yeah it seems to make sense to me given the structure of napalm.  Made one comment inline in the code but it was just a minor thing

napalm_base/__init__.py
dbarrosop: Yes, I want to import `napalm_driver`, not `napalm_driver.driver`. Remember, we discussed about having the load_driver in the `__init__.py` file.

privateip: Makes sense... my point is the line should be either

```return 'napalm_{}.{}'.format(module, module)``` 
or 
 ```return 'napalm_{}'.format(module)```

Drop the extra var :)

dbarrosop: Right! Thanks!

PR #13, opened 2016-04-19
Add get_ntp_peers()

mirceaulinic:
As discussed in https://github.com/napalm-automation/napalm-base/pull/12, will reimplement get_ntp_peers() to return a list of NTP peers as configured on the device.

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -654,6 +654,26 @@ class NetworkDriver(object):
         """
         raise NotImplementedError
 
+
+    def get_ntp_peers(self):
+
+        """
+        Returns a list of NTP peers configured on the device.
+
+        Example::
+
+            [
+                '192.168.0.1',
+                '17.72.148.53',
+                '37.187.56.220',
+                '162.158.20.18'
+            ]
+
+        """
+
+        raise NotImplementedError
+
+
     def get_ntp_stats(self):
 
         """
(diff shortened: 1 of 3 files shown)


mirceaulinic:
@dbarrosop, @ktbyers does ```show ntp assoc``` on IOS returns the complete addresses for remote peers? (Please try with IPv6 -- on JunOS and EOS the output is truncated).

napalm_base/base.py
```diff
@@ -654,6 +654,26 @@ def get_arp_table(self):
...
 
+
+    def get_ntp_peers(self):
+
+        """
+        Returns a list of NTP peers configured on the device.
+
+        Example::
+
+            [
```
dbarrosop: Can we do a dictionary instead? Right now it's going to be a bit sad because it's going to be a dictionary with no values:

{               
                '192.168.0.1': {}
                '17.72.148.53': {}
                '37.187.56.220': {}
                '162.158.20.18': {}
}

The reason is that you can stil do `d.get_ntp_peers.keys()` to get the list but doing a dict of dicts will allow us to extend this method to add more data if needed. Like the key used for authentication, or other configuration parameters. What do you think?

mirceaulinic:
Although for the moment I don't see an obvious need, it is possible to need more information around NTP peers. Moreover would not cost very much to transform it into a dictionary.

mirceaulinic:
@dbarrosop I finished updating the drivers, everything should be fine now.

mirceaulinic:
@dbarrosop all clear now :)

PR #14, opened 2016-04-19
Import/Export policy as list

mirceaulinic:
Although JunOS is the only OS supporting multiple import/export policies for BGP/BGP group/BGP neighbour, we need to update the output of all drivers to return a list of policies applied per level.
In that case, Cisco & EOS drivers will return a list of policies containing at most one element 👯

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -423,11 +423,12 @@ class NetworkDriver(object):
         """
         raise NotImplementedError
 
-    def get_bgp_config(self, group = '', neighbor = ''):
+    def get_bgp_config(self, group='', neighbor=''):
         """
         Returns a dictionary containing the BGP configuration.
         Can return either the whole config, either the config only for a group or neighbor.
         Main dictionary keys represent the group name and the values represent a dictionary having the following keys:
+
             * type (string)
             * description (string)
             * apply_groups (string list)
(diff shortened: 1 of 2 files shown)


mirceaulinic:
Updated for all drivers currently supporting get_bgp_config()

dbarrosop:
I am afraid on how this might affect other's people code. People assumes this is a string so changing it to a list might break some people's code. Would it work if we just keep it as it is and just separate policies with a delimiter in the case there are multiple? I am still feeling a bit uncomfortable with the change we did with get_ntp_peeers/get_ntp_stats although that one made a lot of sense ;)

mirceaulinic:
Indeed we can return a CSV string instead of an array. The problem is that right now, on JunOS when multiple policies are applied, returns an array instead on str :)
We can qualify this as a bug and make the necessary changes for JunOS only, sure.

mirceaulinic:
Closing?

dbarrosop:
> Indeed we can return a CSV string instead of an array...

Yeah, I would rather do that. It doesn't involve changing the API, which should always be avoided if possible, and I don't think it's worth the effort if it affects only one vendor.

I am closing this and the referenced PRs, feel free to reopen them if you want to keep discussing this,

mirceaulinic:
No need to reopen, already discussed and agreed :)
Will submit soon the necessary changes.

PR #16, opened 2016-04-19
Define get_users()

mirceaulinic:
Retrieves users configured on the device

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1191,7 +1191,7 @@ class NetworkDriver(object):
 
             * level (int)
             * password (str)
-            * sshkey (list)
+            * sshkeys (list)
 
         The level is an integer between 0 and 15, where 0 is the lowest access and 15 represents full access to the device.
 
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1074,6 +1074,7 @@ class NetworkDriver(object):
         """
         raise NotImplementedError
 
+
     def traceroute(self, destination, source='', ttl=0, timeout=0):
         """
         Executes traceroute on the device and returns a dictionary with the result.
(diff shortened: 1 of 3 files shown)


napalm_base/base.py
```diff
@@ -1009,3 +1009,30 @@ def get_probes_results(self):
...
         """
         raise NotImplementedError
+
+    def get_users(self):
+        """
+        Returns a dictionary with the configured users.
+        The keys of the main dictionary represents the username. The values represent the details of the user,
+        represented by the following keys:
+
+            * level (int)
```
dbarrosop: What if it's a juniper user? Should we just map the default classes to numbers? Something like:

operator:        1
read-only:       9
super-user:      15

Or should we do it the other way around, naming the levels somehow?

mirceaulinic:
Indeed. Due to the fact that most network engineers are mostly used with the level rather than class name, I preferred to have this field and integer (between 0 and 15). On JunOS will define a map between the class name and level.

```python
_JUNOS_CLASS_CISCO_PRIVILEGE_LEVEL_MAP = {
  'super-user': 15,
  'superuser': 15,
  'operator': 5,
  'read-only': 1,
  'unauthorized': 0
}
```

dbarrosop:
:shipit: 

Do you have any implementation already?

mirceaulinic:
Yup :)

mirceaulinic:
@dbarrosop Doing multiple things right now and I did a mess with the NTP peers, can we postpone it till tomorrow?

dbarrosop:
Of course! : )

mirceaulinic:
Meanwhile, would be possible to merge this one and have the necessary
environment prepared?

Le mar. 26 avr. 2016 18:26, David Barroso <notifications@github.com> a
écrit :

dbarrosop:
Done!

mirceaulinic:
Thanks! :)

>

mirceaulinic:
@dbarrosop All setup & ready here.

PR #17, opened 2016-04-19
Define ping() and traceroute()

mirceaulinic:


mirceaulinic:
@dbarrosop and with this I think we have enough work for the next few days.

mirceaulinic:
@dbarrosop I have a couple of question marks around my head regarding ```traceroute```:

- right now considers only one probe result (first) / hop
- should we consider all (default 3)
- increasing the number of probes/hop could provide better statistics, but
- on JunOS you can't change the number of probes/hop, but on IOS-XR it is possible (also on EOS, probably on NXOS too). In such case, should we add one more argument?

mirceaulinic:
@dbarrosop no, actually on NXOS there's no option to specify the number of queries (not even TTL and timeout) :(
Thus EOS and IOS-XR are the only devices to allow this feature.

mirceaulinic:
@dbarrosop Oh well, it's not that hard anyway to convert to multiple probes/hop. If for some reason will not be needed, one can consider only one of the probes.

mirceaulinic:
@dbarrosop sorry for the errors, it looks like for JunOS & IOS-XR I have not copied the includes from our version and I can solve that quickly. On EOS is not yet ready and I will need some more work to consider all probes results from the traceroute output.

mirceaulinic:
@dbarrosop implemented now on JunOS, IOS-XR, NXOS, EOS and Pluribus. Just needs the base adjustment here: https://github.com/napalm-automation/napalm-base/pull/20 before testing.

mirceaulinic:
@dbarrosop All ```traceroute``` implementations fixed now! :)

mirceaulinic:
@dbarrosop please don't forget about this one: https://github.com/napalm-automation/napalm-iosxr/pull/13

PR #21, opened 2016-04-26
find_txt and convert functions in helpers

mirceaulinic:
I find both functions very useful and needed more and more. In order to not define them again and again in each driver class, I think we should include them in the helpers module (actually that was their purpose) and use wherever needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,3 +19,4 @@ deploy:
 
 script:
   - nosetests ./test/unit/TestGetNetworkDriver.py
+  - nosetests ./test/unit/TestHelpers.py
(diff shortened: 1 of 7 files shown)


mirceaulinic:
After merging this PR, I will update all drivers to use the functions from napalm_base.helpers, instead of the local defined static methods.

dbarrosop:
Could we have some unittests? If drivers are going to rely on these we should have tests. Other than that it's a brilliant idea : )

mirceaulinic:
Sure, I will find some time next week to write the tests and make some
cleanups in the code of the drivers i work on mostly.

Le sam. 30 avr. 2016 19:47, David Barroso <notifications@github.com> a
écrit :

mirceaulinic:
@dbarrosop added some tests too.

dbarrosop:
💯

PR #23, opened 2016-04-27
fixes napalm-automation/napalm#234

privateip:
This adds a new function that will try to dynamically import a module
from a Python egg.  This function is called if the import from file system
fails with an ImportError.

### DIFF
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -21,7 +21,7 @@ import inspect
 from base import NetworkDriver
 
 
-def _import_module(name):
+def _import_module_from_fs(name):
     """Import a module into the current runtime environment.
 
     This function will take a full path module name and break it into
@@ -63,6 +63,58 @@ def _import_module(name):
 
     return mod
 
+def _import_module_from_egg(name, path=None):
+    """Import a module from a zipped egg
+
+
+    Import a module from a Python zipped egg that is consistent with
+    PEP302.  This is function is loosely derived from
+    http://stackoverflow.com/questions/28962344/
+
+    :param str name:
+        The name of the module to import
+
+    :returns:
+        The imported Python module
+    """
+    path = path or sys.path
+    for item in path:
+        module = sys.path_importer_cache.get(item)
+        if module:
+            try:
+                obj = module.find_module(name, [item])
+                if obj:
+                    return obj.load_module(name)
+            except ImportError:
+                pass
+    raise ImportError('No module named %s' % name)
+
+def _import_module(name):
+    """Imports a module based on the string name
+
+    This function will import the module specifed by name by first
+    trying to impor the module from the file system.  If the import
+    fails, then this function will attempt to import the module from
+    a Python egg.   If both imports file, the function will raise an
+    ImportError exception.
+
+    :param str name:
+        The full dotted name of the module to import
+
+    :returns:
+        The imported module object
+
+    :raises ImportError:
+        If the specified name could not be imported
+    """
+    try:
+        mod = _import_module_from_fs(name)
+    except ImportError, exc:
+        try:
+            mod = _import_module_from_egg(name)
+        except ImportError:
+            raise exc
+    return mod
 
 def _load_module(name):
     """Attempt to load a module into the current environment.


mirceaulinic:
@privateip @dbarrosop all the following lines hundreds:

```python
def _import_module_from_fs(name):
    """Import a module into the current runtime environment.
    This function will take a full path module name and break it into
    its parts iteratively attempting to import each one.  The function
    will check to be sure that the module hasn't been previously imported.
    .. doctest::
        >>> _import_module('os') #doctest: +ELLIPSIS
        <module 'os' from '...'>
    :param str name:
        The name of the module to import
    :returns:
        The imported Python module
    """
    parts = name.split('.')
    path = None
    module_name = ''
    fhandle = None

    for index, part in enumerate(parts):
        module_name = part if index == 0 else '%s.%s' % (module_name, part)
        path = [path] if path is not None else path

        try:
            fhandle, path, descr = imp.find_module(part, path)
            if module_name in sys.modules:
                # since imp.load_module works like reload, need to be sure not
                # to reload a previously loaded module
                mod = sys.modules[module_name]
            else:
                mod = imp.load_module(module_name, fhandle, path, descr)
        finally:
            # lets be sure to clean up after ourselves
            if fhandle:
                fhandle.close()

    return mod

def _import_module_from_egg(name, path=None):
    """Import a module from a zipped egg
    Import a module from a Python zipped egg that is consistent with
    PEP302.  This is function is loosely derived from
    http://stackoverflow.com/questions/28962344/
    :param str name:
        The name of the module to import
    :returns:
        The imported Python module
    """
    path = path or sys.path
    for item in path:
        module = sys.path_importer_cache.get(item)
        if module:
            try:
                obj = module.find_module(name, [item])
                if obj:
                    return obj.load_module(name)
            except ImportError:
                pass
    raise ImportError('No module named %s' % name)

def _import_module(name):
    """Imports a module based on the string name
    This function will import the module specifed by name by first
    trying to impor the module from the file system.  If the import
    fails, then this function will attempt to import the module from
    a Python egg.   If both imports file, the function will raise an
    ImportError exception.
    :param str name:
        The full dotted name of the module to import
    :returns:
        The imported module object
    :raises ImportError:
        If the specified name could not be imported
    """
    try:
        mod = _import_module_from_fs(name)
    except ImportError, exc:
        try:
            mod = _import_module_from_egg(name)
        except ImportError:
            raise exc
    return mod

def _load_module(name):
    """Attempt to load a module into the current environment.
    This function will load a module from the Python sys.path.  It will
    check to be sure the module wasn't already loaded.  If the module
    was prevsiouly loaded, it will simply return the loaded module and not
    reload it
    .. doctest::
        >>> _load_module('sys')
        <module 'sys' (built-in)>
    :param str name:
        The name of the module to load
    :returns:
        The named Python module
    :raises ImportError:
        If the module could not be loaded
    """
    try:
        mod = None
        mod = sys.modules[name]
    except KeyError:
        try:
            mod = _import_module(name)
        except ImportError:
            raise
    finally:
        if not mod:
            raise
        return mod


def _napalm_namespace(module):
    """Prepend the default namespace onto the module name.
    The function is only applicable in cases where a module
    name is provided without a full path namespace.  In this
    case, the module name is prepended with the device namespace.
    If the module name is a fully qualified namespace, the
    module name is returned.
    .. doctest::
        >>> _napalm_namespace('test')
        'napalm_test.test'
        >>> _napalm_namespace('napalm.test')
        'napalm.test'
    :param str module:
        The name of the module to verify
    :returns:
        The fully qualified module namespace
    """
    if '.' in module:
        return module
    return 'napalm_{}'.format(module)


def _discover_driver_from_module(module):
    """Return the first subclass of ``NetworkDriver`` found in ``module``."""
    driver = None
    for member in inspect.getmembers(module):
        if inspect.isclass(member[1]):
            if issubclass(member[1], NetworkDriver) and member[1] is not NetworkDriver:
                driver = member[1]
                break
    return driver
```

just to load a driver as like we have tons of drivers to manages and we are overwhelmed?

The previous method was simple and very straightforward:

```python
    driver_mapping = {
        'EOS': EOSDriver,
        'ARISTA': EOSDriver,
        'IOS-XR': IOSXRDriver,
        'IOSXR': IOSXRDriver,
        'JUNOS': JunOSDriver,
        'JUNIPER': JunOSDriver,
        'FORTIOS': FortiOSDriver,
        'NXOS': NXOSDriver,
        'IBM': IBMDriver,
        'IOS' : IOSDriver,
        'PLURIBUS': PluribusDriver
    }
    try:
        return driver_mapping[vendor.upper()]
    except KeyError:
        raise Exception('Vendor/OS not supported: %s' % vendor)
```

I know it's really nice to exercise your Python skills, but in my humble opinion it's not really worth it. What for? Do you really see a gain in here? The only thing I see right now (napalm-automation/napalm#234 says that) is that after some waste of time we managed somehow to introduce new bugs.

dbarrosop:
To be fair, from those hundreds of lines more than 50% are actual documentation. Which is actually nice :)

The new method has the nice property that decouples `napalm-base` from the `drivers`. Meaning that if, let's say, some organization wants to write a driver and keep ownership to themselves they can do it as `napalm-base` does load the driver dynamically without having any prior knowledge. It also doesn't require updating if we implement a new driver for the very same reason. However, this is not a requirement, it's just a nice property.

If we can decouple `napalm-base`  from the drivers in an easier way I am all for it.  An idea to simplify `napalm-base` would be to ditch the 'discovery' of the `NetworkDriver` subclass and just require a `load_driver` method on the `__init_.py` file that returns the actual driver. That might get rid of/simplify a few methods while retaining the nice properties we got.

Worst case, if people thinks it's too complex or that python2.x packaging system is too broken for this to work properly we can also revert to the old static behavior.

@ktbyers @fooelisa @jedelman8 Pinging random people to see what they think : )

hurrycane:
Not clear to me why: https://docs.python.org/2/library/importlib.html cannot be used here?

hurrycane:
Also some tests for this functionality might be great too.

privateip:
> Not clear to me why: https://docs.python.org/2/library/importlib.html cannot be used here?

support for py26

dbarrosop:
> support for py26

NAPALM doesn't support py26 so if that's the reason we should be using it.

ktbyers:
Yes, I am not seeing why we need all this complexity?

It seems like we should be able to do dynamic with importlib or \_\_import\_\_ in a much simpler way i.e. keep dynamic discovery using sys.path.

Someone could still swap in their own custom driver in this solution.

mirceaulinic:
For me the following couple of lines just worked:

```python
>>> import inspect
>>> import importlib
>>> from napalm_base.base import NetworkDriver
>>> mod = importlib.import_module('napalm_junos')
>>> for name, obj in inspect.getmembers(mod):
...     if inspect.isclass(obj) and issubclass(obj, NetworkDriver):
...             print name
...
JunOSDriver
```

This is very minimalistic, but of course we can work on it and make it fancier.

What do you think, is this anything better?

privateip:
as i mentioned earlier in this thread, the code is py26 compliant which is still today the most dominant version of python running in most data centers :( .   There is absolutely thing wrong with using importlib and i would say if you dont care about py26 support then you should go for it and simplify the code.   Unfortunately thats one of the pitfalls of not having a publicly documented strategy and direction for a project is not knowing what platforms and version are to be supported.

mirceaulinic:
@privateip who provided you such terrible statistics?
Moreover the purpose of the NAPALM project is to automate networks not to be installed on all possible servers you might have, but only on a single dedicated server (or inside a container in the worst case). And upgrading one server/container from py26 to py27 should not cost very much.

> the code is py26 compliant which is still today the most dominant version of python running in most data centers :(

jedelman8:
I see both sides here...truth of the matter is, there are MANY larger shops where they can't just go from py26 to py27...it's a fact and unfortunate.

mirceaulinic:
Perhaps @privateip is right and indeed it is not specified that NAPALM supports py27 only. @dbarrosop I think we need to make this update in the documentation.
@jedelman8 probably they are, as you claim, although I don't really know what "MANY" means.
But if you really want to achieve the automation goals, you can afford at least creating a container (takes less than 5mins to setup) -- if a server is not possible, install py27 and then NAPALM.

ktbyers:
We need to update the documentation on this...that NAPALM only supports Python2.7. I will submit a PR on that to add a classifier to setup.py. 

I think the above disagreement is due to differences in Ansible's requirements (i.e. inherently distributed and needs to run on a lot of different servers) and NAPALM (inherently centralized). Since NAPALM is centralized, I think we should focus on PY3 and PY2.6.

The case that might be relevant to NAPALM is network environments that force automation to go through a bastion host. I don't think this case is significant enough to support Python2.6.

dbarrosop:
> We need to update the documentation on this...that NAPALM only supports Python2.7. I will submit a PR on that to add a classifier to setup.py.

Thanks!

> The case that might be relevant to NAPALM is network environments that force automation to go through a bastion host. I don't think this case is significant enough to support Python2.6.

I agree, if people wants to use the latest and the greatest they should be able to, at least, upgrade their python version. It's not like you are forced to have one or the other, a system can have all the python version you want. And it's more work for us supporting old and deprecated python versions that it is for them to upgrade python. And the implications are far greater as that adds complex, bug-prone technical debt to the code.

dbarrosop:
As it seems we agree that we want to keep things simple I am going to close this one.

@privateip thanks for the PR, even if we decided to go the easy way I can tell you I learnt a lot about the internals of the python import systems thanks to this PR. From now on I am going to write only in `golang` XD

PR #24, opened 2016-04-28
New implementation for get_network_driver

mirceaulinic:
@dbarrosop @ktbyers @hurrycane @privateip Any thoughts about that?

No documentation yet, but I will provide in case we all agree this looks and feels better.

### DIFF
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -34,13 +34,13 @@ def get_network_driver(module_name):
     """
     Searches for a class derived form the base NAPALM class NetworkDriver in a specific library.
     The library name must repect the following pattern: napalm_[DEVICE_OS].
-    NAPALM community supports a list of devices and provides the corresponding libraries, for full reference
+    NAPALM community supports a list of devices and provides the corresponding libraries; for full reference
     please refer to the `Supported Network Operation Systems`_ paragraph on `Read the Docs`_.
 
     .. _`Supported Network Operation Systems`: http://napalm.readthedocs.io/en/latest/#supported-network-operating-systems
     .. _`Read the Docs`: http://napalm.readthedocs.io/
 
-    :param module_name:         name of the device operating system, or the name of the library.
+    :param module_name:         the name of the device operating system, or the name of the library.
     :return:                    the first class derived from NetworkDriver, found in the library.
     :raise ModuleImportError:   when the library is not installed, or a derived class from NetworkDriver was not found.
 
@@ -50,6 +50,8 @@ def get_network_driver(module_name):
 
         >>> get_network_driver('junos')
         <class 'napalm_junos.junos.JunOSDriver'>
+        >>> get_network_driver('IOS-XR')
+        <class 'napalm_iosxr.iosxr.IOSXRDriver'>
         >>> get_network_driver('napalm_eos')
         <class 'napalm_eos.eos.EOSDriver'>
         >>> get_network_driver('wrong')
@@ -61,9 +63,9 @@ def get_network_driver(module_name):
 
     try:
         module_name = module_name.lower()  # only lowercase allowed
-        module_install_name = module_name
-        if 'napalm_' not in module_name:  # can also request using napalm_[SOMETHING]
-            module_install_name = 'napalm_{name}'.format(name=module_name)
+        module_install_name = module_name.replace('-', '')  # to not raise error when users requests IOS-XR for e.g.
+        if 'napalm_' not in module_install_name:  # can also request using napalm_[SOMETHING]
+            module_install_name = 'napalm_{name}'.format(name=module_install_name)
         module = importlib.import_module(module_install_name)
     except ImportError:
         raise ModuleImportError(
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -14,165 +14,39 @@
 
 """napalm_base package."""
 
-import sys
-import imp
+# Python std lib
 import inspect
+import importlib
 
-from base import NetworkDriver
-
-
-def _import_module(name):
-    """Import a module into the current runtime environment.
-
-    This function will take a full path module name and break it into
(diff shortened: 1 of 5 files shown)


ktbyers:
@mirceaulinic Testing now.

ktbyers:
@mirceaulinic Some feedback.

```python
napalm_base.exceptions.ModuleImportError
```

doesn't exist.

Also:

```python
    except ImportError as ierr:
```

ierr is never used.


Also I think we need to document the install process both using PIP and for python setup.py install (probably both for an individual driver i.e. I only want to use 'EOS' and full). Currently, it is far from self-explanatory how to do the install.

napalm_base/exceptions.py
```diff
@@ -12,6 +12,9 @@
 # License for the specific language governing permissions and limitations under
 # the License.
 
+class ModuleImportError(Exception):
```
mirceaulinic: @ktbyers here is defined ```napalm_base.exceptions.ModuleImportError```. Was not imported properly?

ktbyers:
@mirceaulinic Also can you add the following to the classifiers section of setup.py as part of this PR?

```python
        'Programming Language :: Python :: 2',
        'Programming Language :: Python :: 2.7',
```

mirceaulinic:
>     except ImportError as ierr:
> ierr is never used.
> 
Indeed, it’s not used — and don’t think it provides any useful information.
> Also I think we need to document the install process both using PIP and for python setup.py install (probably both for an individual driver i.e. I only want to use 'EOS' and full). Currently, it is far from self-explanatory how to do the install.
> 
Good point! We need to spend some time on documentation too!

napalm_base/exceptions.py
ktbyers: My error...I missed you also changed the exception file (what I get for trying to test quickly).

mirceaulinic: OK, good news if everything is fine with the imports. Thanks!

ktbyers:
@mirceaulinic It looks like you have some import issues. 

Here is the error I receive when I try to execute some code (and there is no 'ios' library):

```python
Password: 
  File "napalm_ios_test.py", line 77, in <module>
    main()
  File "napalm_ios_test.py", line 18, in main
    driver = get_network_driver('ios')
  File "/home/gituser/VENV/napalm_meta4/local/lib/python2.7/site-packages/napalm_base-0.12.0-py2.7.egg/napalm_base/__init__.py", line 37, in get_network_driver
    raise napalm_base.exceptions.ModuleImportError(
AttributeError: 'module' object has no attribute 'exceptions'
```

It is strange to me that you are trying to 'import napalm_base' inside of \_\_init\_\_

Shouldn't you have:

```python
from napalm_base.exceptions import ModuleImportError
```

And then change all of the namespace references in the __init__.

Anything else we want to make available at the napalm_base package level should also be imported. We might also want to define an '\_\_all\_\_'.

ktbyers:
FYI, the above changes (made locally fixed the issue for me). I now get the exception I expect:

```
napalm_base.exceptions.ModuleImportError: Cannot import "napalm_ios". Is the library installed?
```

mirceaulinic:
Normally should would like that too, as imported module from a library, but doesn't cost anything to import from local modules instead.

mirceaulinic:
@ktbyers I think it's probably because you run form the installation directory and Python imported the ```__init__.py``` instead of the installed ```napalm_base``` library -- local modules have priority over the installed packages. But we leave it like that now, to avoid any further confusions, although some specifications suggest importing using the lib name.

ktbyers:
@mirceaulinic No, I was running from a totally different directory (when testing). The error was coming from the installed package (in site-packages).

mirceaulinic:
Great! Thanks for your feedback @ktbyers!

ktbyers:
No problem...a bit of a strange namespace issue (the above should fix it, however).

dbarrosop:
There are some tests that were supposed to be testing the import system. They seem to be passing which is good. Could you review the tests and see if I missed something when I wrote them? Maybe we can install an additional driver with `python setup.py` during the install phase and add it to the test as well.

mirceaulinic:
@dbarrosop @ktbyers I made some changes in order to test existing and non-existing drivers.

mirceaulinic:
@dbarrosop Still, this in not ready yet, please allow me some time to come back with proper documentation.

napalm_base/__init__.py
```diff
@@ -14,165 +14,46 @@
...
-    This function will take a full path module name and break it into
-    its parts iteratively attempting to import each one.  The function
-    will check to be sure that the module hasn't been previously imported.
 
-    .. doctest::
+def get_network_driver(module_name):
 
-        >>> _import_module('os') #doctest: +ELLIPSIS
-        <module 'os' from '...'>
+    if not (isinstance(module_name, str) and len(module_name) > 0):
```
privateip: test for instance of basestring not str.  If unicode text is based as the module_name arg, your test will fail.

```
>>> test = 'test string'
>>> isinstance(test, str)
True
>>> isinstance(unicode(test), str)
False
>>> isinstance(unicode(test), basestring)
True
```

dbarrosop:
Are you happy with this one? I think I am going to pull this off 0.13.0 and have a dedicated release for this enhancement.

mirceaulinic:
Personally, for me it worked. But I would propose to not hurry and test it more before merging.

mirceaulinic:
And perhaps we can add more features if needed.

dbarrosop:
Agreed, that's why I moved it to a dedicated release.

dbarrosop:
Are you happy with this one? If you are let me know and I will review it again.

mirceaulinic:
I tested it (actually I put it in production already) and seems to be fine so far. I am just thinking that we can make it more flexible when identifying the driver (although it's not a real need right now). Otherwise, I don't have anything else to add, only to ask you to test it yourself too :)

Issue #6, opened 2016-03-06
Document load_template


dbarrosop:

mirceaulinic:
I was thinking that we could add one extra param: ```eternal_template``` and let the users pass an absolute path to a non-standardized template.  What do you think, @dbarrosop?

dbarrosop:
I think that you were reading my mind. Was planning to suggest that as well :)

How do you envision this though? Given the name you gave to the variable I think we were thinking in two different implementations. The way I thought about this was as follows:

1. I have my own folder `PATH` where I put all my templates.
2. Each driver has it's own subfolder, ie, `PATH/eos`, `PATH/junos`, etc...
3. When I want to use my own template I just do `d.load_template('my_service', template_path=PATH)`

So my code doesn't have to figure out the current driver or anything. Is this what you were thinking?

mirceaulinic:
Yes, exactly! I don't have a schedule it right now, but will make some time probably next week and implement it.

mirceaulinic:
@dbarrosop I think this one can be closed by now

PR #27, opened 2016-05-05
Load_template: load from custom path

mirceaulinic:
Regarding the discussion with @dbarrosop in #6.

load_template is documented in the base class, but if necessary, we can copy it here too. Or we can define it in the helper function and load the docstring dynamically in the method :P

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -85,15 +85,17 @@ class NetworkDriver(object):
         """
         Will load a templated configuration on the device.
 
-        :param cls: instance of the driver class
-        :param template_name: identifies the template name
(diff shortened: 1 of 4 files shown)


mirceaulinic:
@dbarrosop: how would you feel about a custom template passed as optional arg? I think it would be nice too!

dbarrosop:
What do you mean? As a string to `load_template`? What would be the benefit vs just going with `load_merge` in that case?

mirceaulinic:
Yes, as string. Basically the only benefit would be that napalm renders the template for you, without having the need to install Jinja on the local machine (assuming that you make the calls to a remote server).

PR #36, opened 2016-05-12
Add get_ntp_servers()

mirceaulinic:
Similar to ```get_ntp_peers```, retrieves the configuration of the NTP servers.

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -671,7 +671,6 @@ class NetworkDriver(object):
         """
         raise NotImplementedError
 
-
     def get_ntp_peers(self):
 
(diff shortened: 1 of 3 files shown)


mirceaulinic:
@dbarrosop Any thoughts on this one?

dbarrosop:
Oh, I set the label but forgot to reply... Looks good to me, simple and to the point : ) I will merge this one when we have an implementation at least.

mirceaulinic:
I have the code for a couple of drivers already prepared (I think even before I opened this PR), I just need some time to organise my repos and will try to come soon with something.

PR #41, opened 2016-06-17
Skip tests if NotImplementedError is thrown.

jzohrab:
Currently, Napalm tests have many NotImplementedErrors due to devices not having all features.  As a dev and longtime TDD guy, I'm accustomed to running my tests, and having them either pass or fail.  I find that false negatives clutter my test runs and interrupt my dev "flow", eg:

```
(venv) ~/Documents/Projects/napalm-nxos$ python -m unittest test.unit.TestNXOSDriver
.E.EE..E......EEE..E....
... lots of (intentional) error noise ...
```

This PR checks for NotImplementedErrors for each method call, and if that error is raised, it skips the test.  Here's the same test run after the patch is applied:

```
(venv) ~/Documents/Projects/napalm-nxos$ python -m unittest test.unit.TestNXOSDriver
.s.ss..s......sss..s....
----------------------------------------------------------------------
Ran 24 tests in 0.046s

OK (skipped=8)
```

You can see what has been skipped with the verbose flag:

```
(venv) ~/Documents/Projects/napalm-nxos$ python -m unittest --verbose test.unit.TestNXOSDriver 2>&1 | grep skipped
test_get_bgp_config (test.unit.TestNXOSDriver.TestGetterNXOSDriver) ... skipped ''
test_get_bgp_neighbors_detail (test.unit.TestNXOSDriver.TestGetterNXOSDriver) ... skipped ''
test_get_environment (test.unit.TestNXOSDriver.TestGetterNXOSDriver) ... skipped ''
test_get_interfaces_counters (test.unit.TestNXOSDriver.TestGetterNXOSDriver) ... skipped ''
test_get_probes_config (test.unit.TestNXOSDriver.TestGetterNXOSDriver) ... skipped ''
test_get_probes_results (test.unit.TestNXOSDriver.TestGetterNXOSDriver) ... skipped ''
test_get_route_to (test.unit.TestNXOSDriver.TestGetterNXOSDriver) ... skipped ''
test_ping (test.unit.TestNXOSDriver.TestGetterNXOSDriver) ... skipped ''
```

The Python language designers actually discussed special handling for NotImplementedError in http://bugs.python.org/issue1466, but that feature isn't implemented.  I don't really want to subclass the test runner, this change I propose is much simpler.

I think this would simplify testing for all projects as a whole.  Discussion welcome!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,3 +20,4 @@ deploy:
 script:
   - nosetests ./test/unit/TestGetNetworkDriver.py
   - nosetests ./test/unit/TestHelpers.py
+  - nosetests ./test/unit/TestNapalmTestFramework.py
--- a/test/unit/TestNapalmTestFramework.py
+++ b/test/unit/TestNapalmTestFramework.py
@@ -1,0 +1,17 @@
+import unittest
+import napalm_base.helpers
+from napalm_base.base import NetworkDriver
+import napalm_base.test.base as ntb
+
+# Note: don't import the TestGettersNetworkDriver class itself, or
+# nose finds it and tries to run the tests for it, which yields
+# errors.
+
+class TestSkipNotImplemented(unittest.TestCase, ntb.TestGettersNetworkDriver):
+    """Ensure that any tests are skipped if not implemented."""
+
+    def setUp(self):
+        class FakeThing(NetworkDriver):
+            def __init__(self):
+                pass
+        self.device = FakeThing()
--- a/napalm_base/test/base.py
+++ b/napalm_base/test/base.py
@@ -17,7 +17,7 @@ from __future__ import print_function
 from napalm_base import exceptions
 import difflib
 import models
-
+from unittest import SkipTest
 
 class TestConfigNetworkDriver:
 
@@ -38,8 +38,11 @@ class TestConfigNetworkDriver:
             print(line)
 
     def test_replacing_and_committing_config(self):
-        self.device.load_replace_candidate(filename='%s/new_good.conf' % self.vendor)
-        self.device.commit_config()
+        try:
+            self.device.load_replace_candidate(filename='%s/new_good.conf' % self.vendor)
+            self.device.commit_config()
+        except NotImplementedError:
+            raise SkipTest()
 
         # The diff should be empty as the configuration has been committed already
         diff = self.device.compare_config()
@@ -55,6 +58,8 @@ class TestConfigNetworkDriver:
         try:
             self.device.load_replace_candidate(filename='%s/new_typo.conf' % self.vendor)
             self.device.commit_config()
+        except NotImplementedError:
+            raise SkipTest()
         except exceptions.ReplaceConfigException:
             self.device.load_replace_candidate(filename='%s/initial.conf' % self.vendor)
             diff = self.device.compare_config()
@@ -153,12 +158,18 @@ class TestGettersNetworkDriver:
         return correct_class and same_keys
 
     def test_get_facts(self):
-        facts = self.device.get_facts()
+        try:
+            facts = self.device.get_facts()
+        except NotImplementedError:
+            raise SkipTest()
         result = self._test_model(models.facts, facts)
         self.assertTrue(result)
 
     def test_get_interfaces(self):
-        get_interfaces = self.device.get_interfaces()
+        try:
+            get_interfaces = self.device.get_interfaces()
+        except NotImplementedError:
+            raise SkipTest()
         result = len(get_interfaces) > 0
 
         for interface, interface_data in get_interfaces.iteritems():
@@ -167,7 +178,10 @@ class TestGettersNetworkDriver:
         self.assertTrue(result)
 
     def test_get_lldp_neighbors(self):
-        get_lldp_neighbors = self.device.get_lldp_neighbors()
+        try:
+            get_lldp_neighbors = self.device.get_lldp_neighbors()
+        except NotImplementedError:
+            raise SkipTest()
         result = len(get_lldp_neighbors) > 0
 
         for interface, neighbor_list in get_lldp_neighbors.iteritems():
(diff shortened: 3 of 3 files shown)


napalm_base/test/test_NapalmTestFramework.py
```diff
@@ -0,0 +1,14 @@
+import unittest
+import napalm_base.helpers
+from napalm_base.base import NetworkDriver
+from napalm_base.test.base import TestConfigNetworkDriver, TestGettersNetworkDriver
+
+
+class TestGetterNetworkDriver_passes(unittest.TestCase, TestGettersNetworkDriver):
```
mirceaulinic: Where do you expect this to be used and how?

mirceaulinic:
The tests are mainly executed by Travis, but for sure this is very useful when running them locally!

napalm_base/test/test_NapalmTestFramework.py
jzohrab: This test class was mostly a sanity check for me only to ensure that the skip covered all cases.  Plus I like to add tests for pretty much everything to ensure I have a good understanding of how things fit together.

I really don't expect this test to be run much, maybe add it to Travis, but it's kind of silly on its own. :-)

jzohrab:
I prefer having a test suite that I can run on my desktop, as much as possible.  Should be doable with fakes, even if a dev doesn't have a full set of devices to test against.

napalm_base/test/test_NapalmTestFramework.py
dbarrosop: What's the benefit of this? Getters are mocked already, so you can just run the test without this `FakeThing` as there is a fake class already on getter test.

jzohrab: I wasn't aware.  I read through https://github.com/napalm-automation/napalm-base/blob/develop/napalm_base/test/base.py and couldn't see an existing fake. Since the TestGettersNetworkDriver doesn't inherit from TestCase, it's not run as a test, so I created this test by copying the outline from one of the implementations and reducing it to this.  Am I looking in the wrong spot?

If there is already a test somewhere that calls all the methods in TestGettersNetworkDriver with a fake object then of course this is redundant.  If not, including this test would be good for completeness -- e.g., if a new method is added to the base driver, and its test doesn't handle NotImplementedError, then this test would catch that.

Really, it's a test of the tests, which is sometimes good for project stability.  It's not a problem for me if you would prefer to remove it or disable it, I'm presenting it as an option only.

dbarrosop: > if a new method is added to the base driver, and its test doesn't handle NotImplementedError, then this test would catch that

I was trying to understand the motivation and that one seems like a good one for : ) Could you enable it in `.travis.yml`?

jzohrab: Before I do that, is this the right location and name for this file?  It's in `napalm_base/test/test_NapalmTestFramework.py`.  I could move this to be part of the `base.py` file, which might be a better location for it, and would be right next to the TestGetter class itself.

dbarrosop: Use `/napalm-base/test/unit/TestNapalmTestFramework.py` and name the class `TestSkipNotImplemented`.

jzohrab:
Incorporated notes from @dbarrosop, added to Travis.  Runs OK locally:

```
$ python -m unittest test.unit.TestNapalmTestFramework
sssssssssssssssssssss
...[snip]...
OK (skipped=21)
```

jzohrab:
Travis failed with `TestGettersNetworkDriver instance has no attribute 'device'`.  I'm running the tests successfully with unittest, I guess there is something different in the setup.  Investigating.

At the end of this will likely rebase and squash, remove the commit history noise.

jzohrab:
Rebased, as commits were messy due to Travis failures.  The current code works with both unittest and nose:

```
$ python -m unittest test.unit.TestNapalmTestFramework
sssssssssssssssssssss
...
OK (skipped=21)
$ nosetests ./test/unit/TestNapalmTestFramework.py
SSSSSSSSSSSSSSSSSSSSS
...
OK (SKIP=21)
```

jzohrab:
Nice, passed.

@dbarrosop, this PR may also simplify the .travis.yml in all the projects.  eg, in the Nexus project, rather than having to spell out all the tests you want to include in its [.travis.yml](https://github.com/napalm-automation/napalm-nxos/blob/develop/.travis.yml), you should be able to specify the full test class.

.travis.yml
```diff
@@ -20,3 +20,4 @@ deploy:
 script:
   - nosetests ./test/unit/TestGetNetworkDriver.py
   - nosetests ./test/unit/TestHelpers.py
+  - nosetests ./test/unit/TestNapalmTestFramework.py
```
mirceaulinic: ```LF``` perhaps?

mirceaulinic: Anyway maybe this test should be removed?

test/unit/TestNapalmTestFramework.py
```diff
@@ -0,0 +1,17 @@
+import unittest
+import napalm_base.helpers
+from napalm_base.base import NetworkDriver
+import napalm_base.test.base as ntb
+
+# Note: don't import the TestGettersNetworkDriver class itself, or
+# nose finds it and tries to run the tests for it, which yields
+# errors.
+
+class TestSkipNotImplemented(unittest.TestCase, ntb.TestGettersNetworkDriver):
```
mirceaulinic: I assume this will be removed from the PR, am I right?

jzohrab: @mirceaulinic - I was thinking it would stay, per the discussion with David in https://github.com/napalm-automation/napalm-base/pull/41#discussion-diff-67585869.  If a new method is added to the base driver, and its test doesn't handle NotImplementedError, then this test would catch that.

Ideally, the NotImplementedError will be used to simplify the .travis.yml files in each project, and so it would need to continue to work going forward.

mirceaulinic: Right, did not see this discussion, sorry.

jzohrab: NP, clearly the comment "Ensure that any tests are skipped if not implemented" was insufficient, but that was the intent ... I could add more notes if you think that would improve clarity.  Say "yes please" and I'll do it.

mirceaulinic: This is fine, it is clear enough. It was just my bad I did not read carefully.

mirceaulinic:
🏆

PR #42, opened 2016-06-22
New testing framework

dbarrosop:
This is the new testing framework. It's based on py.test this time because I think it's way more interesting and powerful than the rest. This is how it works:

1. To mock tests create a folder with the test name. Inside create as many folders as you want with different name, each one containing different mocked data. The result will be that for each subfolder a different test case will be generated. See https://github.com/dbarrosop/napalm-eos/tree/test_framework/test/unit/mocked_data
1. If there is no test folder that means is used by the `open` function.
1. I have grouped functions that we were rewriting all the time. For example, how to read a json or a text file. See https://github.com/dbarrosop/napalm-base/blob/test_framework/napalm_base/test/double.py
1. Instead of editing files, to control wether to run the test with mocked data or what is the username, password, etc., we can use env variables. See https://github.com/dbarrosop/napalm-base/blob/test_framework/napalm_base/test/conftest.py#L6
1. The `NotImplemented` check for the tests is done with a decorator. The decorator does more stuff as well, like setting some parameters to figure out the test case. See https://github.com/dbarrosop/napalm-base/blob/test_framework/napalm_base/test/getters.py#L13

This is the result of an execution:

```
py.test test/unit/test_getters.py -v
==================================================================== test session starts =====================================================================
platform darwin -- Python 2.7.11, pytest-2.9.1, py-1.4.31, pluggy-0.3.1 -- /Users/dbarroso/.virtualenvs/napalm/bin/python2.7
cachedir: .cache
rootdir: /Users/dbarroso/workspace/napalm/napalm-eos, inifile: setup.cfg
plugins: pylama-7.0.7, pythonpath-0.7
collecting 0 items
[NAPALM] We failed to find test cases for test_get_bgp_neighbors. Please, add test cases in /Users/dbarroso/workspace/napalm/napalm-eos/test/unit/mocked_data/test_get_bgp_neighbors
collected 5 items

test/unit/test_getters.py::TestGetter::test_get_bgp_neighbors_detail[empty_case] <- ../napalm-base/napalm_base/test/getters.py SKIPPED
test/unit/test_getters.py::TestGetter::test_get_interfaces[normal] <- ../napalm-base/napalm_base/test/getters.py PASSED
test/unit/test_getters.py::TestGetter::test_get_bgp_neighbors[test_case0] <- ../napalm-base/napalm_base/test/getters.py SKIPPED
test/unit/test_getters.py::TestGetter::test_get_facts[something] <- ../napalm-base/napalm_base/test/getters.py PASSED
test/unit/test_getters.py::TestGetter::test_get_facts[something_else] <- ../napalm-base/napalm_base/test/getters.py PASSED

============================================================ 3 passed, 2 skipped in 0.01 seconds =============================================================
```

Test `test/unit/test_getters.py::TestGetter::test_get_bgp_neighbors_detail[empty_case]` is skipped because there the method returns a `NotImplemented` exception and `test/unit/test_getters.py::TestGetter::test_get_bgp_neighbors` because there is not test case defined. Not that there is a log entry notifying about that:

`[NAPALM] We failed to find test cases for test_get_bgp_neighbors. Please, add test cases in /Users/dbarroso/workspace/napalm/napalm-eos/test/unit/mocked_data/test_get_bgp_neighbors`

See napalm-automation/napalm-eos#25 for details on how to implement this on a driver.

### DIFF
--- a/napalm_base/test/double.py
+++ b/napalm_base/test/double.py
@@ -1,0 +1,46 @@
+"""Base class for Test doubles."""
+
+import json
+import re
+import os
+import sys
+
+
+class BaseTestDouble:
+    """Base class for test doubles."""
+
+    def __init__(self, *args, **kwargs):
+        """Initiate object."""
+        self.current_test = ''
+        self.current_test_case = ''
+
+    def find_file(self, filename):
+        """Find the necessary file for the given test case."""
+        # Find base_dir of submodule
+        module_dir = os.path.dirname(sys.modules[self.__module__].__file__)
+
+        full_path = os.path.join(module_dir, 'mocked_data',
+                                 self.current_test, self.current_test_case, filename)
+
+        if os.path.exists(full_path):
+            return full_path
+        else:
+            raise Exception("Couldn't find file with mocked data: {}".format(full_path))
+
+    @staticmethod
+    def sanitize_text(text):
+        """Remove some weird characters from text, useful for building filenames from commands."""
+        regexp = r'[\[\]\*\^\+\s\|\/]'
+        return re.sub(regexp, '_', text)[0:150]
+
+    @staticmethod
+    def read_json_file(filename):
+        """Parse a json file and return its content."""
+        with open(filename) as data_file:
+            return json.load(data_file)
+
+    @staticmethod
+    def read_txt_file(filename):
+        """Return the content of a file."""
+        with open(filename) as data_file:
+            return data_file.read()
--- a/napalm_base/test/double.py
+++ b/napalm_base/test/double.py
@@ -25,7 +25,7 @@ class BaseTestDouble:
         if os.path.exists(full_path):
             return full_path
         else:
-            raise Exception("Couldn't find file with mocked data: {}".format(full_path))
+            raise IOError("Couldn't find file with mocked data: {}".format(full_path))
 
     @staticmethod
     def sanitize_text(text):
@@ -44,3 +44,14 @@ class BaseTestDouble:
         """Return the content of a file."""
         with open(filename) as data_file:
             return data_file.read()
+
+    @property
+    def expected_result(self):
+        """Return the expected result for the current test case."""
+        filename = self.find_file('expected_result.json')
+
+        with open(filename, mode='r') as f:
+            try:
+                return json.loads(f.read())
+            except ValueError:
+                raise ValueError("No JSON object could be decoded on filename: {}".format(filename))
--- a/napalm_base/test/double.py
+++ b/napalm_base/test/double.py
@@ -55,3 +55,20 @@ class BaseTestDouble:
                 return json.loads(f.read())
             except ValueError:
                 raise ValueError("No JSON object could be decoded on filename: {}".format(filename))
+
+
+def _string_key_to_int(param):
+    """For a given dictionary, convert all strings that represent a number into an int."""
+    new_dict = {}
+
+    if isinstance(param, list):
+        return [_string_key_to_int(element) for element in param]
+    elif isinstance(param, dict):
+        for key, value in param.items():
+            try:
+                new_dict[int(key)] = _string_key_to_int(value)
+            except ValueError:
+                new_dict[key] = _string_key_to_int(value)
+        return new_dict
+    else:
+        return param
--- a/napalm_base/test/getters.py
+++ b/napalm_base/test/getters.py
@@ -19,14 +19,14 @@ def wrap_test_cases(func):
 
         try:
             result = func(cls)
-            not_implented = False
+            not_implemented = False
         except NotImplementedError:
-            not_implented = True
+            not_implemented = True
 
         cls.device.device.current_test = ''
         cls.device.device.current_test_case = ''
 
-        if not_implented:
+        if not_implemented:
             pytest.skip("Method not implemented")
         else:
             return result
--- a/napalm_base/test/getters.py
+++ b/napalm_base/test/getters.py
@@ -2,22 +2,33 @@
 
 from __future__ import print_function
 
-import json
-
-import models
-import pytest
-
 import functools
+import json
+import itertools
+
+from double import BaseTestDouble
 
 import helpers
 
-from double import BaseTestDouble
+import models
+
+
+import pytest
+
+
+def list_dicts_diff(prv, nxt):
+    """Compare two lists of dicts."""
+    result = []
+    for prv_element, nxt_element in itertools.izip_longest(prv, nxt, fillvalue={}):
+        intermediate_result = dict_diff(prv_element, nxt_element)
+        if intermediate_result:
+            result.append(intermediate_result)
+    return result
 
 
 def dict_diff(prv, nxt):
     """Return a dict of keys that differ with another config object."""
     keys = set(prv.keys() + nxt.keys())
-
     result = {}
 
     for k in keys:
@@ -49,12 +60,13 @@ def wrap_test_cases(func):
             not_implemented = False
 
             if isinstance(cls.device.device, BaseTestDouble):
-                diff = dict_diff(result, cls.device.device.expected_result)
-                try:
-                    assert not diff, "Expected result varies on some keys {}".format(diff)
-                except AssertionError:
+                if isinstance(result, list):
+                    diff = list_dicts_diff(result, cls.device.device.expected_result)
+                else:
+                    diff = dict_diff(result, cls.device.device.expected_result)
+                if diff:
                     print("Resulting JSON object was: {}".format(json.dumps(result)))
-                    raise
+                    raise AssertionError("Expected result varies on some keys {}".format(diff))
 
         except NotImplementedError:
             not_implemented = True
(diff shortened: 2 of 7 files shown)


dbarrosop:
Pinging @jzohrab and @mirceaulinic

napalm_base/test/getters.py
```diff
@@ -0,0 +1,91 @@
...
+def wrap_test_cases(func):
+    """Wrap test cases."""
+    @functools.wraps(func)
+    def wrapper(cls, test_case):
+        cls.device.device.current_test = func.__name__
+        cls.device.device.current_test_case = test_case
+
+        try:
+            result = func(cls)
+            not_implented = False
```
jzohrab: Intentional typo?  May as well spell out the full var name. :-)

dbarrosop: An unintentional one. Will fix it later in case you spot more : )

jzohrab:
This looks interesting, partitioning the test data by folder makes sense.

Is there a way to have "expected outputs" beside the expected inputs as well?  Currently the tests check the validity of the structure only, but not the returned data, which would be nice.  This may be an idea for a future iteration.

I've not run tests using actual devices, I've only run them with the fakes.  Using an environment variable to toggle is much better than editing code.

Do you have an example done for an existing driver, even if it's incomplete?

dbarrosop:
> Is there a way to have "expected outputs"

It should be doable to add an extra assert on each method to check some expected output per driver/test case. I will implement it before merging this PR.

dbarrosop:
@jzohrab @mirceaulinic 

More progress on this, I think it should be ready by now. Not sure if we are missing something here. The latest addition is mostly that for mocked tests we have to provide an expected output as well. The check is done in the wrapper function so current tests require almost no changes whatsoever to support this. It's mostly the mocked data that has to be adapted. Here is an example of running the tests right now with several test cases:

```
(napalm) ➜  napalm-eos git:(test_framework) py.test test/unit/test_getters.py -v
============================================================== test session starts ===============================================================
platform darwin -- Python 2.7.11, pytest-2.9.1, py-1.4.31, pluggy-0.3.1 -- /Users/dbarroso/.virtualenvs/napalm/bin/python2.7
cachedir: .cache
rootdir: /Users/dbarroso/workspace/napalm/napalm-eos, inifile: setup.cfg
plugins: pylama-7.0.7, pythonpath-0.7
collecting 0 items
[NAPALM] We failed to find test cases for test_get_bgp_neighbors_detail. Please, add test cases in /Users/dbarroso/workspace/napalm/napalm-eos/test/unit/mocked_data/test_get_bgp_neighbors_detail
collected 6 items

test/unit/test_getters.py::TestGetter::test_get_bgp_neighbors_detail[test_case0] <- ../napalm-base/napalm_base/test/getters.py SKIPPED
test/unit/test_getters.py::TestGetter::test_get_interfaces[normal] <- ../napalm-base/napalm_base/test/getters.py PASSED
test/unit/test_getters.py::TestGetter::test_get_interfaces[sub_key_doesnt_match] <- ../napalm-base/napalm_base/test/getters.py FAILED
test/unit/test_getters.py::TestGetter::test_get_bgp_neighbors[common] <- ../napalm-base/napalm_base/test/getters.py PASSED
test/unit/test_getters.py::TestGetter::test_get_facts[common] <- ../napalm-base/napalm_base/test/getters.py PASSED
test/unit/test_getters.py::TestGetter::test_get_facts[this_one_fails] <- ../napalm-base/napalm_base/test/getters.py FAILED

==================================================================== FAILURES ====================================================================
______________________________________________ TestGetter.test_get_interfaces[sub_key_doesnt_match] ______________________________________________

cls = <test_getters.TestGetter instance at 0x10bd51c68>, test_case = 'sub_key_doesnt_match'

    @functools.wraps(func)
    def wrapper(cls, test_case):
        cls.device.device.current_test = func.__name__
        cls.device.device.current_test_case = test_case

        try:
            result = func(cls)
            not_implemented = False

            if isinstance(cls.device.device, BaseTestDouble):
                diff = dict_diff(result, cls.device.device.expected_result)
                try:
>                   assert not diff, "Expected result varies on some keys {}".format(diff)
E                   AssertionError: Expected result varies on some keys {u'Management1': {'is_enabled': {'expected': False, 'result': True}}, u'Ethernet1': {'description': {'expected': None, 'result': u''}}}

../napalm-base/napalm_base/test/getters.py:54: AssertionError
-------------------------------------------------------------- Captured stdout call --------------------------------------------------------------
Resulting JSON object was: {"Ethernet2": {"is_enabled": true, "description": "", "last_flapped": 1466586841.4151127, "is_up": true, "mac_address": "08:00:27:10:c4:8f", "speed": 0}, "Management1": {"is_enabled": true, "description": "", "last_flapped": 1466586841.4284112, "is_up": true, "mac_address": "08:00:27:20:b9:04", "speed": 1000}, "Ethernet1": {"is_enabled": true, "description": "", "last_flapped": 1466586841.4148579, "is_up": true, "mac_address": "08:00:27:c6:00:f0", "speed": 0}, "Ethernet4": {"is_enabled": true, "description": "", "last_flapped": 1466586841.415464, "is_up": true, "mac_address": "08:00:27:e0:12:d2", "speed": 0}, "Ethernet3": {"is_enabled": true, "description": "", "last_flapped": 1466586841.4152465, "is_up": true, "mac_address": "08:00:27:1f:60:43", "speed": 0}}
___________________________________________________ TestGetter.test_get_facts[this_one_fails] ____________________________________________________

cls = <test_getters.TestGetter instance at 0x10bcd52d8>, test_case = 'this_one_fails'

    @functools.wraps(func)
    def wrapper(cls, test_case):
        cls.device.device.current_test = func.__name__
        cls.device.device.current_test_case = test_case

        try:
            result = func(cls)
            not_implemented = False

            if isinstance(cls.device.device, BaseTestDouble):
                diff = dict_diff(result, cls.device.device.expected_result)
                try:
>                   assert not diff, "Expected result varies on some keys {}".format(diff)
E                   AssertionError: Expected result varies on some keys {'os_version': {'expected': u'4.15.5M-3054042.4155M', 'result': u'4.15.4M'}}

../napalm-base/napalm_base/test/getters.py:54: AssertionError
-------------------------------------------------------------- Captured stdout call --------------------------------------------------------------
Resulting JSON object was: {"os_version": "4.15.4M", "uptime": 877650, "interface_list": ["Ethernet1", "Ethernet2", "Ethernet3", "Ethernet4", "Management1"], "vendor": "Arista", "serial_number": "", "model": "vEOS", "hostname": "localhost", "fqdn": "localhost"}
================================================= 2 failed, 3 passed, 1 skipped in 0.02 seconds ==================================================
```

As you can see:

1. It complains there is no test case for `test_get_bgp_neighbors_detail`.
2. Skips `test_get_bgp_neighbors_detail` because it's not implemented.
3. For `TestGetter.test_get_interfaces[sub_key_doesnt_match]` it complains that a couple of subkeys don't match (one has a different value and the other doesnt exist)
4. `TestGetter.test_get_facts[this_one_fails]` reports one of the keys dont match.

What do you think? Is this something we should migrate to? If you agree I will finish migrating the tests on `napalm-base` and fix `napalm-eos`. Then I would appreciate if you could help out migrating other drivers as well.

dbarrosop:
BTW, before the merge I will try to document how this works and I will improve the docstrings. I know the code for the testing framework is extremely confusing right now as it relies on some magic and `py.test` features : )

jzohrab:
@dbarrosop - that "expected_results" idea is a great add, checking the actual values will add a lot of value, I think.

Is the "expected_results" check required, or can it be skipped?  For example, you may want to do a smoketest/sanity check and not worry about results -- but maybe that's a different kind of test, and not a unit test.

Also, fyi, in my unit tests I reused some files in different tests.  Maybe for this style of testing we could use simlinks.

jzohrab:
Re the json diffs: I created a small method in another project to help me diff with testing with large strings which might be useful for this.  Can submit a PR later.  https://github.com/jzohrab/capirca/blob/add_yml_parser/tests/integration/filediff.py

dbarrosop:
> Is the "expected_results" check required, or can it be skipped?

It's required for the mocked data. If you are running against a real device they are skipped. You can also make a value "dynamic" by saying that the expected value is `"..."`. For example, EOS reports the timestampo of when it was booted, that means that when we calculate the `uptime` the value changes on every run even on the mocked tests so I had to implement this dynamism : )

https://github.com/napalm-automation/napalm-eos/pull/25/files#diff-44ef7e3d2a70a5dfd77657d49d2a879eR14

> Also, fyi, in my unit tests I reused some files in different tests. Maybe for this style of testing we could use simlinks.

Being text files I would just copy them. I would prefer to avoid weird interaction between unix-like systems, github, windows... We already had some issues for those super long filenames with non-ascii characters

> Re the json diffs

That's cool. Might be useful in the future. For this particular case, comparing json/dict objects as strings has the inconvenience that they are, by definition unsorted, so two identical objects might be different as a string because keys are on a different order.

jzohrab:
Thanks, good points.

Ack on copying vs simlinks, I go back and forth on that myself.  Ppl at my current job seem to like using simlinks so I think I'm getting converted.  Copying is a good standard to choose.

We could work around the json diffing, I've done something similar in the past.  This is a proposal only, we can keep in in mind and add it later if it would help people out during test implementation.

dbarrosop:
> We could work around the json diffing

What do you mean? The diff is already implemented by the method `dict_diff` in the `getters.py` file? Or is there something else you might want to check?

jzohrab:
I didn't read through all the code, it's possible that everything is already covered and that the test feedback is already extremely clear.  Once it gets in and I start working with it, I'll have a better idea, and should have held off on the idea until then.  :-)

mirceaulinic:
Although proposed pretty long time ago, only today I had time to read and understand how it works. I don't have anything to add, I like it, let's go for it!

napalm_base/test/double.py
```diff
@@ -0,0 +1,74 @@
...
+
+        if os.path.exists(full_path):
+            return full_path
+        else:
+            raise IOError("Couldn't find file with mocked data: {}".format(full_path))
+
+    @staticmethod
+    def sanitize_text(text):
+        """Remove some weird characters from text, useful for building filenames from commands."""
+        regexp = r'[\[\]\*\^\+\s\|\/]'
```
mirceaulinic: When passing some commands to IOS-XR or other Cisco platforms that contain parts of config eventually, they might contain also weird characters like ", $, = or !

dbarrosop: Added!

PR #48, opened 2016-07-04
Test connectivity script

mirceaulinic:
In my proposal https://github.com/napalm-automation/napalm/pull/267, @jzohrab suggested that a script to test the connectivity would be very handy.
I propose a script called ```cl_napalm_test```, reusing some parts of the code from ```cl_napalm_configure```.

### DIFF
--- a/napalm_base/clitools/cl_napalm_configure.py
+++ b/napalm_base/clitools/cl_napalm_configure.py
@@ -10,16 +10,11 @@ import logging
 logger = logging.getLogger('cl-napalm-config.py')
 
 
-def build_help():
+def build_help(connect_test=False):
     parser = argparse.ArgumentParser(
         description='Command line tool to handle configuration on devices using NAPALM.'
                     'The script will print the diff on the screen',
         epilog='Automate all the things!!!'
-    )
-    parser.add_argument(
-        dest='config_file',
-        action='store',
-        help='File containing the configuration you want to deploy.'
     )
     parser.add_argument(
         dest='hostname',
@@ -48,25 +43,10 @@ def build_help():
         help='Host Operating System.'
     )
     parser.add_argument(
-        '--strategy', '-s',
-        dest='strategy',
-        action='store',
-        choices=['replace', 'merge'],
-        default='replace',
-        help='Strategy to use to deploy configuration. Default: replace.'
-    )
-    parser.add_argument(
         '--optional_args', '-o',
         dest='optional_args',
         action='store',
         help='String with comma separated key=value pairs that will be passed via optional_args to the driver.',
-    )
-    parser.add_argument(
-        '--dry-run', '-d',
-        dest='dry_run',
-        action='store_true',
-        default=None,
-        help='Only returns diff, it does not deploy the configuration.',
     )
     parser.add_argument(
         '--debug',
@@ -74,6 +54,27 @@ def build_help():
         action='store_true',
         help='Enables debug mode; more verbosity.'
     )
+    if not connect_test:
+        parser.add_argument(
+            '--strategy', '-s',
+            dest='strategy',
+            action='store',
+            choices=['replace', 'merge'],
+            default='replace',
+            help='Strategy to use to deploy configuration. Default: replace.'
+        )
+        parser.add_argument(
+            dest='config_file',
+            action='store',
+            help='File containing the configuration you want to deploy.'
+        )
+        parser.add_argument(
+            '--dry-run', '-d',
+            dest='dry_run',
+            action='store_true',
+            default=None,
+            help='Only returns diff, it does not deploy the configuration.',
+        )
     args = parser.parse_args()
 
     if args.password is None:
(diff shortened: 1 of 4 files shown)


jzohrab:
This looks good.  We should add it to the docs.  I have a [quickstart branch](https://github.com/napalm-automation/napalm/tree/quickstart) in napalm, which is still work-in-progress, and we could add a note to there, or on the README somewhere.  How is this script run?

mirceaulinic:
Sure, my idea is to add it in the README as the very first step in the short list of checks to be followed (see #267 in the main repo). I encourage your to include it in the quickstart too!
Regarding the syntax, there's nothing special, it's just like the config script, but without three params: config_file, strategy and dry_run: 
```bash
cl_napalm_test -v<vendor> -u <user> -p <password> -o <optional_args> <hostname> --debug
```

dbarrosop:
Great idea. However, I am not a fan of having a script depend so much of another script. Might lead to reusability issues or even breaking stuff if we forget that other scripts depend on this other script. Comments inline.

napalm_base/clitools/cl_napalm_configure.py
```diff
@@ -10,18 +10,13 @@
 logger = logging.getLogger('cl-napalm-config.py')
 
 
-def build_help():
+def build_help(connect_test=False):
```
dbarrosop: I'd rather have a `help.py' or similar file with the common help and use it as a starting point for all the scripts. That way all the scripts will have a common import and no script will depend on another script.

napalm_base/clitools/cl_napalm_configure.py
```diff
@@ -95,11 +96,8 @@ def configure_logging(debug):
     logger.addHandler(ch)
 
 
-class CustomException(Exception):
-    pass
+def open_connection(vendor, hostname, user, password, optional_args):
```
dbarrosop: I am not sure about this method. You can just write the test script as:

    with driver(hostname, user, password, optional_args=optional_args) as device:
        print('Successfully connected to the device.')

So there isn't really any reusability here and the context manager of `napalm` checks if there is an exception that is not handled by napalm and asks the user to open a github issue:

https://github.com/napalm-automation/napalm-base/blob/develop/napalm_base/base.py#L44

mirceaulinic:
> I am not a fan of having a script depend so much of another script.

Me neither, I just thought that _maybe_ we won't need others in the future and I preferred applying small changes... but who knows? Will move the common parts in a helper module then.

napalm_base/clitools/cl_napalm_configure.py
mirceaulinic: Agreed.

napalm_base/clitools/cl_napalm_configure.py
mirceaulinic: I am aware of those features.
```open_connection``` is used in both ```cl_napalm_configure and ```cl_napalm_test``` - in the second is enough just open the connection using a ```with as ``` block indeed. But in ```cl_napalm_configure``` is not, as the object ```driver``` is needed afterwards to load the config. If inside the function ```open_connection``` if we'd have a ```with as ``` block, the connection will be closed at the end of the block.

File: napalm_base/clitools/helpers.py [hunk 3]
```diff
@@ -0,0 +1,124 @@
...
+        logger.setLevel(logging.INFO)
+
+    ch = logging.StreamHandler(sys.stdout)
+    formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+    ch.setFormatter(formatter)
+    logger.addHandler(ch)
+    return logger
+
+
+def open_connection(logger, vendor, hostname, user, password, optional_args):
```
dbarrosop: I am still not sold on this method. We lose the benefits of the context manager and the whole method is actually two lines:

```
optional_args = json.loads(optional_args)
with driver(hostname, user, password, optional_args=optional_args) as device:
   #code goes here
```

What was the reasoning of this method again? I think I might have missed something here.

Note that I am not arguing the implementation of this method, I am arguing if we need it or not. I personally think this method is unnecessary.

mirceaulinic: It was just because I wanted to reuse some parts of the code. But sure, I can leave as is in ```cl_napalm_conigure``` and write pretty much the same ```with as ```` block in  ```cl_napalm_test``` - it's anyway very tiny.

dbarrosop: Yeah, it's just two lines and I like using the context manager as it tries to "capture" exceptions and report if we are missing some.

mirceaulinic: Hum... I'm afraid it's more than two redundant lines:

```python
logger.debug('Getting driver for OS "{driver}"'.format(driver=vendor))
driver = get_network_driver(vendor)

if optional_args is not None:
    optional_args = {x.split('=')[0]: x.split('=')[1] for x in optional_args.replace(' ', '').split(',')}

logger.debug('Connecting to device "{device}" with user "{user}" and optional_args={optional_args}'.format(
                    device=hostname, user=user, optional_args=optional_args))
with driver(hostname, user, password, optional_args=optional_args) as device:
```

But if you really want, I'm fine with that.

mirceaulinic: Removed the helper, but personally I don't like redundant code.

Issue #40, opened 2016-06-17
Add peer router id into get_bgp_neighbors_detail()


inetuid:
At present the remote peer router id is only available in the output of get_bgp_neighbors() and not in get_bgp_neighbors_detail().

@dbarrosop suggested that get_bgp_neighbors_detail()[my_vrf][my_neigh_ip] could be used to get this information (slight typo I think, probably should be get_bgp_neighbors()[my_vrf][my_neigh_ip]) but that supposes that you have the VRF information available (or else you have to loop through the VRFs to find the peer).

If 'routing_table' in get_bgp_neighbors_detail() is the same as the VRF in get_bgp_neighbors() then this isn't an issue. If they are not, then I propose adding the remote peer router id into the get_bgp_neighbors_detail() method.

mirceaulinic:
I can add the peer router id information for all devices supporting ```get_bgp_neighbors_detail``, no problem.

dbarrosop:
> but that supposes that you have the VRF information available (or else you have to loop through the VRFs to find the peer).

How else are you supposed to do it? You need to know what you are looking for or loop through the data. I actually realized that `get_bgp_neighbors_detail` and `get_bgp_config` are flawed by design as I just realized they don't support VRF. We will have to revisit that.

inetuid:
Not sure if the docs reflect reality, but for get_bgp_neighbors() the docs say 

> The keys for the first dictionary will be the vrf

so that suggests its VRF aware.
.

mirceaulinic:
Indeed ```get_bgp_neighbors``` does, but ```get_bgp_neighbors_detail``` does not, although it should. But this will be probably fixed soon and both will have VRF support!

mirceaulinic:
Thanks @inetuid for raising this issue!

dbarrosop:
Can we close this one?

mirceaulinic:
Yes, the latest release contains these details now. We can close this issue.

PR #49, opened 2016-07-06
Structure adjust for get_bgp_neighbors_detail

mirceaulinic:
Proposal to add the vrf key in the layer 1 dictionary of the output for ```get_bgp_neighbors_detail```.
Will fix https://github.com/napalm-automation/napalm-junos/issues/28 too.
I think we can take this chance to add the ```router_id```: see https://github.com/napalm-automation/napalm-base/issues/40

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -742,11 +742,13 @@ class NetworkDriver(object):
 
         :param neighbor_address: Retuns the statistics for a spcific BGP neighbor.
 
-        The keys of the dictionary represent the AS number of the neighbors.
-        Inner dictionaries contain the following fields:
+        Returns a dictionary of dictionaries. The keys for the first dictionary will be the vrf (global if no vrf).
+        The keys of the inner dictionary represent the AS number of the neighbors.
+        Leaf dictionaries contain the following fields:
             * up (True/False)
             * local_as (int)
             * remote_as (int)
+            * router_id (string)
             * local_address (string)
             * routing_table (string)
             * local_address_configured (True/False)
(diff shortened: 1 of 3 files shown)


mirceaulinic:
@dbarrosop Thoughts?

dbarrosop:
Looks good as per the previous discussion we had. I was wondering though if we should make it look more like the `get_bgp_neighbors` method though.

In that case it would be more like:

```
{
global: {
   peers: {
       '1.2.3.4': {data goes here}
       '1.2.3.5': {data goes here}
    }
 }
}
```

It's a much bigger change than we initially discussed but it would make the output more consistent with the rest of the methods. I am fine if you prefer to keep your proposal. Just think about it and let me know what you prefer : )

mirceaulinic:
I have mixed feelings about that.
Leaving aside that I completely missed the VRF key, the design was intended to be slightly different than ```get_bgp_neighbors```, grouping the peers by the remote AS number - I find it more natural.
So I would leave the AS key. What I would be tempted to do is replacing the list with a dict - but that's just a detail, isn't it?
Also I am afraid to make two changes at a time, I learned that changing concomitantly two details is not a very good idea.
For the moment I would say to pop in only the VRF key to not cause too much damage. Afterwards, if we'll _really_ want them more consistent we can reconsider this - but I doubt since the data is the same and one can whenever play a bit and easily rearrange.

dbarrosop:
I guess it feels more natural because you are thinking in your own use case. If you think from an object/model perspective the remote-as it's just an attribute from a peering relation : ) Using the ASN doesn't help at all if you think in other types of networks like a L3 datacenter where you might have several dozens of relations and all of them have the same remote-AS. In any case, I agree with you that now it's "too late" to change it. Let's make the transition as easy as possible. In the future I would like to accommodate the OC model anyway.

mirceaulinic:
OK, I see you merged it, I will do my best to come ASAP with the fix for all our current implementations. Probably later today or tomorrow.
Totally agree with your arguments. Will do it like that when implementing the OC mode, especially when it's much easier now to discuss design perspectives as all take place here, in napalm-base.

PR #51, opened 2016-07-07
advertised_prefix_count key

### LINKED ISSUE #50: advertise_prefix_count or advertised_prefix_count in get_bgp_neighbor_detail()
The napalm_base/test/models.py uses advertise_prefix_count whereas the documentation (http://napalm.readthedocs.io/en/latest/base.html) uses both. It would seem more natural for it to be advertised_prefix_count to match received/accepted/suppressed etc but it would be good to confirm which it is.

mirceaulinic:
Fixes #50.
Nice catch @inetuid!

### DIFF
--- a/napalm_base/test/base.py
+++ b/napalm_base/test/base.py
@@ -286,7 +286,7 @@ class TestGettersNetworkDriver:
         for vrf, vrf_ases in get_bgp_neighbors_detail.iteritems():
             result = result and isinstance(vrf, unicode)
             for remote_as, neighbor_list in vrf_ases.iteritems():
-                result = result and isinstance(remote_as, unicode)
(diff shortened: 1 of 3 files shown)


napalm_base/test/base.py
```diff
@@ -286,7 +286,7 @@ def test_get_bgp_neighbors_detail(self):
         for vrf, vrf_ases in get_bgp_neighbors_detail.iteritems():
             result = result and isinstance(vrf, unicode)
             for remote_as, neighbor_list in vrf_ases.iteritems():
-                result = result and isinstance(remote_as, unicode)
+                result = result and isinstance(remote_as, int)
```
mirceaulinic: Mistake from my previous PR. Sorry I had to add this here, but it's important!

PR #54, opened 2016-07-08
New helper: mac

mirceaulinic:
I think it's time to have a standard in the format of the MAC Addresses.
Easy to use:

```python
from napalm_base.helper import mac
mac('0123.4567.89ab')
```

### DIFF
--- a/napalm_base/helpers.py
+++ b/napalm_base/helpers.py
@@ -7,11 +7,29 @@ import sys
 # third party libs
 import jinja2
 import textfsm
+from netaddr import EUI
+from netaddr import mac_unix
+from netaddr.core import AddrFormatError
 
 # local modules
 import napalm_base.exceptions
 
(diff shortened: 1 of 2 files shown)


dbarrosop:
Merging this to test the dependent PRs. We should add unittests and document them or they will just go unnoticed.

dbarrosop:
There are a couple of tests failing. I guess we can ignore in this PR the `get_bgp_neighbors_detail` one as there is another PR dealing with changes there.

mirceaulinic:
> We should add unittests and document them or they will just go unnoticed.

Right, very good point. Not sure though where would be the right place to add them - perhaps defining a new section in http://napalm.readthedocs.io/en/latest/support/index.html? In the same time I'm thinking the target are the developers not users

dbarrosop:
> n the same time I'm thinking the target are the developers not users

Agreed, we need a developers section.

PR #58, opened 2016-07-20
Added proposal for 'get_optics' method

narJH27:


dbarrosop:
This is very nice : )

I still have to find out a few details about how to best integrate openconfig with napalm but maybe the way to go is merging your code and, when we have a better way of integrating OC we can just reuse your code for a new method that does all the OC magic. Nevermind, don't worry about that :)

In addition to what's in here:

1. When you have a vendor implementation, please, link it to this PR.
2. We will have to add this method here: https://github.com/napalm-automation/napalm/blob/develop/docs/support/index.rst#getters-support-matrix

narJH27:
Thanks @dbarrosop ! I will make sure to link every vendor implementation that I touch for this method, to this PR. 

Would you be able to add the method to the matrix or do you want me to do that?

dbarrosop:
> Would you be able to add the method to the matrix or do you want me to do that?

You can do it yourself if you want. I have no idea which drivers you are implementing it on so not sure how to fill the table :P

mirceaulinic:
I would have added few other fields, some of them being:

- temperature
- voltage
- bias current

But we can add them later, no worries.

napalm_base/base.py
```diff
@@ -1436,3 +1436,68 @@ def get_users(self):
...
+                                * max (float)
+                            * laser_bias_current
+                                * instant (float)
+                                * avg (float)
+                                * min (float)
+                                * max (float)
+
+        Example:
+
+            {
```
mirceaulinic: Personally I do not find this example very intuitive since I need to look in the structure above and compare.
```physical_channels``` and ```channel``` are supposed to be integers therefore we'll need to replace the keys below with some random values (0).

Comment (mirceaulinic) [inline napalm_base/base.py]: Also you closed too many brackets :P

File: napalm_base/test/base.py [hunk 2]
```diff
@@ -465,3 +465,24 @@ def test_get_users(self):
...
+
+        try:
+            get_optics = self.device.get_optics()
+        except NotImplementedError:
+            raise SkipTest()
+
+        result = isinstance(get_optics, dict)
+
+        for intf, intf_optic in get_optics.iteritems():
+            for phy_chann in intf_optic:
```
mirceaulinic: ```phy_chan``` maybe?

PR #59, opened 2016-07-20
Fixed test for get_optics

narJH27:


mirceaulinic:
Rather than changing the base tests because https://github.com/napalm-automation/napalm-eos/pull/32 is failing, why not leave the output as dict and adapt the output from napalm-eos? I find it more elegant to have it indexed by the interface name.

narJH27:
The only issue with adopting that approach is that, IOSXR does not conform to the same parameters. JUNOS does. @dbarrosop your thoughts?

napalm_base/test/base.py
```diff
@@ -473,14 +473,14 @@ def test_get_optics(self):
...
 
-        result = isinstance(get_optics, dict)
+        result = isinstance(get_optics, list)
 
-        for intf, intf_optic in get_optics.iteritems():
-            for phy_chann in intf_optic:
-                for chan in intf_optic[phy_chan]:
-                    for item in intf_optic[phy_chan][chan]:
+        for item in get_optics:
+            for phy_chann in item:
```
mirceaulinic: ```phy_chan```

mirceaulinic:
What do you mean? You can't get the interface name? In your test I do not see any variable that might play the role of the interface name. Can you provide a proper example of the output expected?

mirceaulinic:
And what's the XML request on IOS-XR to retrieve the optics details?

napalm_base/test/base.py
narJH27: yes that is a typo. will be fixed!

dbarrosop:
I would certainly leave the interface as the entry key to the returning object and the channel as a list, which is how it's documented in `napalm-base` right now. Unless I am misinterpreting something.

narJH27:
`show controllers dwdm <interface number> optic` is the command on XR. There is no global command on XR that gives details about all the optics at once, like in JUNOS or EOS.

To answer your question about the interface name, yes we get the interface name for sure. However, since the data structure returned by `get_optics()` is a multi-level dict with the interface name being the top level key, I wasnt sure how to model the test such that I can validate the interface name too. Any thoughts?

mirceaulinic:
> Any thoughts?

Yes, I think you should leave the output unchanged. Adapt only the example and the tests - currently you output is designed to have a depth of 6 levels, while in your tests you go up to 5 levels - you missed something:


```python
>>> get_optics = {
...   'et1': {
...     0: {
...         0: [
...             {
...               'index': 0,
...               'state': {
...                   'input_power': {
...                       'instant': 0.0,
...                       'avg': 0.0,
...                       'min': 0.0,
...                       'max': 0.0,
...                   },
...                   'output_power': {
...                       'instant': 0.0,
...                       'avg': 0.0,
...                       'min': 0.0,
...                       'max': 0.0,
...                   },
...                   'laser_bias_current': {
...                       'instant': 0.0,
...                       'avg': 0.0,
...                       'min': 0.0,
...                       'max': 0.0,
...                   }
...               }
...             }
...         ]
...     }
...   }
... }
>>>
>>> for intf, intf_optic in get_optics.iteritems():
...    for phy_chan in intf_optic:
...        for chan in intf_optic[phy_chan]:
...            for item in intf_optic[phy_chan][chan]:
...                for state, state_vals in (
...                        intf_optic[phy_chan][chan][item].iteritems()):
...                             print state_vals
...
  File "<stdin>", line 6, in <module>
TypeError: list indices must be integers, not dict
```

mirceaulinic:
You probably wanted to do this:

```python
>>> for intf, intf_optic in get_optics.iteritems():
...   for phy_chan, chan in intf_optic.iteritems():
...     for stuff, more_stuff in chan.iteritems():
...       for item in more_stuff:
...          for state, state_vals in item.iteritems():
...             print state_vals
...
0
{'output_power': {'max': 0.0, 'avg': 0.0, 'instant': 0.0, 'min': 0.0}, 'laser_bias_current': {'max': 0.0, 'avg': 0.0, 'instant': 0.0, 'min': 0.0}, 'input_power': {'max': 0.0, 'avg': 0.0, 'instant': 0.0, 'min': 0.0}}
```

mirceaulinic:
@narJH27 

> show controllers dwdm <interface number> optic is the command on XR. There is no global command on XR that gives details about all the optics at once, like in JUNOS or EOS.

Please do not use CLI commands where there's XML request equivalent.

Login into your IOS-XR box, and type: ```xml echo format``` then paste the following lines:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Request MajorVersion="1" MinorVersion="0">
  <Get><Operational><DWDM/></Operational></Get>
</Request>
```

napalm_base/test/base.py
```diff
@@ -473,14 +473,13 @@ def test_get_optics(self):
...
+        result = isinstance(get_optics, list)
 
         for intf, intf_optic in get_optics.iteritems():
-            for phy_chann in intf_optic:
-                for chan in intf_optic[phy_chan]:
-                    for item in intf_optic[phy_chan][chan]:
-                        for state, state_vals in (
-                                intf_optic[phy_chan][chan][item].iteritems()):
+            for phy_chan, chan in intf_optic.iteritems():
+                for stuff, more_stuff in chan.iteritems():
```
mirceaulinic: Please do not leave ```stuff``` and ```more_stuff``` :)

narJH27:
Yep will use the XML equivalents for IOSXR.

mirceaulinic:
Only need to update the Example -- see my comment in https://github.com/napalm-automation/napalm-base/pull/58/files#r71615447 then rebase everything in one commit and we're done here

narJH27:
@dbarrosop could you please merge this PR when you get a chance? I am waiting on this so that I can push my changes to the EOS driver, so that the build will not fail

dbarrosop:
@mirceaulinic the output you are using in your example is wrong. The example in the dosctring is correct:

https://github.com/napalm-automation/napalm-base/blob/develop/napalm_base/base.py#L1468

napalm_base/base.py
```diff
@@ -1469,6 +1469,8 @@ def get_optics(self):
 
             {
                     'et1': {
+                        0: {
```
dbarrosop: What is this? This and next line shouldn't be here.

mirceaulinic:
@dbarrosop 
> The example in the dosctring is correct:

Does ```'physical_channels'``` or ```'channel'``` look like valid ```int``` as specified few lines above? https://github.com/napalm-automation/napalm-base/blob/develop/napalm_base/base.py#L1448

napalm_base/base.py
mirceaulinic: It'a valid example for the ```int``` field ```physical_channels``` maybe?

narJH27:
@dbarrosop Not sure what you mean by, 'do the `physical_channels` and `channel` look like int'? The example just lists the default value of phy_chan and chan for an interface. Can you explain?

dbarrosop:
@mirceaulinic, you are right, the description is not clear enough but the example is good. The docstring could be somehting like:

```
        Returns a dictionary where the keys are as listed below:
            * intf_name (unicode)
                * physical_channels (key_name)
                    * channels (list of dicts)
                        * index (int)
                        * state
                            * input_power
                                * instant (float)
                                * avg (float)
                                * min (float)
                                * max (float)
                            * output_power
                                * instant (float)
                                * avg (float)
                                * min (float)
                                * max (float)
                            * laser_bias_current
                                * instant (float)
                                * avg (float)
                                * min (float)
                                * max (float)
        Example:
            {
                    'et1': {
                        'physical_channels': {
                            'channel': [
                                {
                                    'index': 0,
                                    'state': {
                                        'input_power': {
                                            'instant': 0.0,
                                            'avg': 0.0,
                                            'min': 0.0,
                                            'max': 0.0,
                                        },
                                        'output_power': {
                                            'instant': 0.0,
                                            'avg': 0.0,
                                            'min': 0.0,
                                            'max': 0.0,
                                        },
                                        'laser_bias_current': {
                                            'instant': 0.0,
                                            'avg': 0.0,
                                            'min': 0.0,
                                            'max': 0.0,
                                        },
                                    }
                                }
                            ]
                        }
                    }
                }
            }
```

Does that make more sense?

narJH27:
@dbarrosop So in this example, what would the value for `physical_channels` and `channel` be in the default case? will it take a value of 0?

dbarrosop:
It's actually hard to describe, we should use YANG xD

`physical_channels` is a key name which returns a dictionary.
`channel` is a key name as well that returns a list of dicts

If we think in OOP `physical_channels` and `channel` would be attributes:

```
result['et1'].physical_channels.channel[0].index # this would return 0
result['et1'].physical_channels.channel[0].state.input_power. instant # this would return 0.0
```

Does this make more sense?

mirceaulinic:
@dbarrosop Yesterday, comparing the structure and the example, I thought the same. But afterwards, after comparing with the implementation in https://github.com/napalm-automation/napalm-eos/pull/32, seemed that the fields in subject are ```int```.
I would like to as @narJH27 to make it clear if possible. Can you please paste, for example the output from your eos implementation to seem what you actually had in mind? Thanks!

dbarrosop:
Maybe we should just write a class to make it more clear and testable?

narJH27:
@mirceaulinic My implementation currently uses '0' as the default value for these, thereby making them ints.

However, since @dbarrosop mentioned that this will be modeled off of OC, I am unsure how the keys will have to be managed in the scheme of this returned data structure. @dbarrosop when you say, write a class and make it testable, what exactly do you mean?

dbarrosop:
I will send later today some code that I think it will help

dbarrosop:
So I rewrote your test:

```
result = device.get_optics()
for iface, iface_data in result.iteritems():
    assert isinstance(iface, str)
    for channel in iface_data['physical_channels']['channel']:
        assert len(channel) == 2
        assert isinstance(channel['index'], int)
        for field in ['input_power', 'output_power', 'laser_bias_current']:
            assert len(channel['state'][field]) ==  4
            assert isinstance(channel['state'][field]['instant'], float)
            assert isinstance(channel['state'][field]['avg'], float)
            assert isinstance(channel['state'][field]['min'], float)
            assert isinstance(channel['state'][field]['max'], float)
```

I hope this helps clarifying the structure.

narJH27:
Just wanted to confirm if the resulting data structure will be a list of dicts, or a dict of dicts?

dbarrosop:
It's a dict where the key is the interface name. If you try a list of dicts with the test I pasted you in my previous comment you will get an error.

dbarrosop:
There is a problem with your current tests and is that it's easy to forge a dictionary that will pass the test without actually being compliant. Could you remove the model you previously added from the `models.py` and replace the test with the code of my previous comment? That code should deal with "rogue" keys as well : )

Thanks!

narJH27:
Done.

narJH27:
@mirceaulinic @dbarrosop  Quick question for you guys! 

I am trying to retrieve the light levels for IOSXR using `<Get><Operational><DWDM/></Operational></Get>`.

However, the output of the above rpc-command, screws up the interface names to dwdm0/0/0/0 etc and so on. I would actually like to retrieve the output of the command `show controllers <intf-name> phy` instead. I have tried scouting through the XML schemas guide for XR and have not found anything that can help out here. Any thoughts on what can be done as an alternative?

PR #60, opened 2016-07-22
Fixed bug in test_get_optics()

narJH27:


narJH27:
@mirceaulinic @dbarrosop Repeating my question since #59 is now closed and I dint want you guys to miss it. 

I am trying to retrieve the light levels for IOSXR using <Get><Operational><DWDM/></Operational></Get>.

However, the output of the above rpc-command, screws up the interface names to dwdm0/0/0/0 etc and so on. I would actually like to retrieve the output of the command show controllers <intf-name> phy instead. I have tried scouting through the XML schemas guide for XR and have not found anything that can help out here. Any thoughts on what can be done as an alternative?

dbarrosop:
> However, the output of the above rpc-command, screws up the interface names to dwdm0/0/0/0 etc 

Could you elaborate on that one, please? I don't have an ASR so I am not sure what you mean with that :)

narJH27:
Ok. So the RPC command `<Get><Operational><DWDM/></Operational></Get>` provides a detailed output of the dwdm interfaces on the ASR. However, in the output, for instance, the interface names are specified as `dwdm0/0/0/0` as opposed to `TenGigE0/0/0/0`. Now I am not sure how to validate the actual interface name due to the output of the RPC command.

Instead, if there were a way to retrieve the output provided by `show controllers <intf> physical` using an XML RPC, it would really help. I was hoping to have a discussion about this and figure out the way forward.

mirceaulinic:
@narJH27 Please open an issue in napalm-iosxr and I will have a look next week.

PR #67, opened 2016-09-11
Added missing requirement

dbarrosop:
Seems like `netaddr` was missing from the `requirements.txt` file.

Will release a quick fix once this is approved.


ktbyers:
+1

Issue #65, opened 2016-09-09
add get_config()


dbarrosop:
Original ticket napalm-automation/napalm#291 created by @grizz
### Add a function to get the full config as a text blob

Would obviously be platform specific, but would be handy for using NAPALM for development and device changes.

dbarrosop:
The description of the method can be found here:

https://github.com/napalm-automation/napalm-base/pull/69/files

Thoughts?

dbarrosop:
@grizz let me know if this proposal looks good to you.

grizz:
Looks good, thanks for stubbing it in. I wonder if it would be better to pass a flag to determine which configs to get since it can take a while on some platforms?

mirceaulinic:
This has been implemented on all drivers, except IBM - which we consider deprecated.
So this is fixed IMO.

PR #69, opened 2016-09-11
Added get_config method

dbarrosop:
Related to napalm-automation/napalm-base#65

@napalm-automation/council thoughts? Will send EOS implementation as a POC.

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1517,11 +1517,16 @@ class NetworkDriver(object):
         """
         raise NotImplementedError
 
-    def get_config(self):
+    def get_config(self, retrieve=None):
         """
         Return the configuration of a device.
 
-        The object returned is a dictionary with the following keys:
+        Args:
+            retrieve(string): Which configuration type you want to populate, default is all of them.
+                The rest will be set to "".
+
+        Returns:
+          The object returned is a dictionary with the following keys:
             - running(string) - Representation of the native running configuration
             - candidate(string) - Representation of the native candidate configuration. If the
               device doesnt differentiate between running and startup configuration this will an
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1516,3 +1516,15 @@ class NetworkDriver(object):
                 }
         """
         raise NotImplementedError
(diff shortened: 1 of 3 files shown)


mirceaulinic:
This looks good and makes sense to me.
Two minor points:
- Given that a more standardised protocol (e.g. NETCONF) defines three conceptual config DBs, I would also add candidate_config as a key - whose value would be empty string when not in configure config mode.
- Tiny detail - but I always love tiny details: can we rename the keys ```running```, ```startup``` and eventually ```candidate``` as it is pretty clear they are about config?

It is a very quick implementation I can probably come with some PRs tomorrow.

dbarrosop:
Makes sense, I added `candidate` and shortened the names.

dbarrosop:
I have also implemented the change to napalm-eos napalm-automation/napalm-eos#38

grizz:
Added to junos at https://github.com/grizz/napalm-junos/commit/f373b1603c84dcdfa1ce875417d3c49215e6b492 - will do PR once the design is set

dbarrosop:
I added a "retrieve" argument to select which config you want in case you want only one. I also added some tests to make sure it's honored.

ktbyers:
@dbarrosop You might want to change the default of retrieve. My first thought was retrieve meant to we use cached data versus pulling the configs (obviously it doesn't from looking at your code)

IMO, it would make more sense to have retrieve default to 'all' or something like that (as opposed to None).

dbarrosop:
Makes sense

dbarrosop:
I changed both base and eos.

grizz:
@dbarrosop looks good to me, I thought None was better for all, but either works :)

dbarrosop:
Yeah, from a coding point of view None makes more sense but I understand it's confusing from a user perspective.

mirceaulinic:
Hi @grizz can you open a pull request with the JunOS implementation you have?

grizz:
@mirceaulinic sure, I'm not seeing docs on how to run the tests and provide mock data, could you point me in the right direction?

PR #74, opened 2016-09-13
Add find_version to napalm_base

ktbyers:
Added find_version() to both the setup.py file for napalm_base and to helpers.py.

Other packages should just be able to add the following into their setup.py

```
from napalm_base.helpers import find_version
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,33 +1,14 @@
 """setup.py file."""
-
 import uuid
-import os
-import re
 
 from setuptools import setup, find_packages
 from pip.req import parse_requirements
+from napalm_base.helpers import find_version
 
 __author__ = 'David Barroso <dbarrosop@dravetech.com>'
 
 install_reqs = parse_requirements('requirements.txt', session=uuid.uuid1())
 reqs = [str(ir.req) for ir in install_reqs]
-
-def find_version(*file_paths):
-    """
-    This pattern was modeled on a method from the Python Packaging User Guide:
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -10,9 +12,26 @@
 install_reqs = parse_requirements('requirements.txt', session=uuid.uuid1())
 reqs = [str(ir.req) for ir in install_reqs]
 
+def find_version(*file_paths):
```
mirceaulinic: Why is not used the ```find_version``` from the helpers instead?

ktbyers: Yeah, your right...updated so find_version is only in napalm_base.helpers

setup.py
```diff
@@ -1,9 +1,9 @@
 """setup.py file."""
-
 import uuid
 
 from setuptools import setup, find_packages
 from pip.req import parse_requirements
+from napalm_base.helpers import find_version
```
mirceaulinic: I did not try - but are you sure it is possible to import a module from the package you want to install in the setup file? I would rather import relatively as ```.napalm_base.helpers```. Perhaps it works like that too - will check it out tomorrow unless you can confirm meanwhile.

ktbyers: That was my original concern...but I tested it and it worked fine.

napalm_base.helpers should always be on the directory path relative to setup.py. I guess it might not work if you are not in that current directory...I will test that part shortly.

PR #87, opened 2016-09-17
pylama cleanup of napalm-base

ktbyers:
This should be all pylama cleanup. Please test, however.

### DIFF
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -39,19 +39,22 @@ if not(sys.version_info.major == 2 and sys.version_info.minor == 7):
 
 
 def get_network_driver(module_name):
-
     """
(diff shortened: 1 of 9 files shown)


ktbyers:
Also should make sure I didn't break any of the auto-generated documentation (built from the docstrings).

dbarrosop:
This is great! Will do some tests : )

Could you add `pylama .` at the end of the `script` section in the file `.travis.yml`?

ktbyers:
@dbarrosop It is not done, but I put a dent in it...

Note, I am ignoring all the openconfig/pyangbind pylint warnings.

dbarrosop:
Ok, I will merge it right after the new testing framework. I will try to make it pass then : )

PR #89, opened 2016-09-17
adding coveralls

fooelisa:
Added coveralls integration to see if this is something we would like to have.


mirceaulinic:
https://github.com/napalm-automation/napalm-base/pull/91 will probably increase with 1-2% the coverage :P

dbarrosop:
I really want this, howerver, bear in mind that for us, functional/acceptance tests are more important than unit tests and coverage.

fooelisa:
Can't seem to be able to move this to pytest instead of nose, it doesn't pick up the tests at all so it seems. Could you take a look @dbarrosop, what's missing here?

ogenstad:
Try: `py.test --cov-report term-missing test/` in .travis.yml

fooelisa:
No `missing-term` will change the report as to displaying missing lines or not, however pytest is not actually finding/starting the tests at all here (which it does in eos). I think they have to be called "test_*" to be picked up.

ogenstad:
Ah ok. Yes pytest needs the tests to begin with test_

.travis.yml
```diff
@@ -5,7 +5,9 @@ python:
 
 install:
   - pip install .
+  - pip install -r requirements-dev.txt
   - pip install -r test/unit/requirements.txt
+  - pip install coveralls
```
dbarrosop: Why don't you just add it to `requirements-dev.txt`?

.travis.yml
```diff
@@ -17,6 +19,7 @@ deploy:
     branch: master
 
 script:
-  - nosetests ./test/unit/TestGetNetworkDriver.py
-  - nosetests ./test/unit/TestHelpers.py
-  - nosetests ./test/unit/TestNapalmTestFramework.py
+  - py.test --cov-report= test/
+  # - pylama .
```
dbarrosop: Why are commenting this one?

.travis.yml
fooelisa: No reason, would you prefer it there?

.travis.yml
fooelisa: leftover from my testing, pylama was producing a lot of output for base

fooelisa:
oh now I remember, pylama was erroring, did you see this in other modules? -eos looks set up the same to me 

`napalm_base/__init__.py WARNING: __all__ is defined as a list, this means pydocstyle cannot reliably detect contents of the __all__ variable, because it can be mutated. Change __all__ to be an (immutable) tuple, to remove this warning. Note, pydocstyle uses __all__ to detect which definitions are public, to warn if public definitions are missing docstrings. If __all__ is a (mutable) list, pydocstyle cannot reliably assume its contents. pydocstyle will proceed assuming __all__ is not mutated.`

.travis.yml
dbarrosop: Yes, please

dbarrosop:
This has been broken for a while, closing it. Please, reopen and fix when you have the time.

PR #90, opened 2016-09-17
Update helpers.py

BobbyGR:
NXOS:
napalm-automation/napalm-nxos#20

Not sure if this was still needed, but this does it raw without needing multiple imports.  Takes any 12 alphanum chars and turns it into the mac format requested.

### DIFF
--- a/napalm_base/helpers.py
+++ b/napalm_base/helpers.py
@@ -8,8 +8,6 @@ import re
 # third party libs
 import jinja2
 import textfsm
-from netaddr import EUI
-from netaddr import mac_unix
 from netaddr import IPAddress
 from netaddr.core import AddrFormatError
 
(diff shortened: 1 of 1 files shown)


BobbyGR:
self.assertEqual(napalm_base.helpers.mac('fake'), '')
AssertionError: 'Missing MAC Charcaters' != ''

will just return the results as this is a thing?

mirceaulinic:
@BobbyGR what issues do you have with the current implementation?

With your proposal an address like ```29:91:4d:f0::``` does not look valid, although it is exactly what a Cisco device returns.

```python
>>> mac('29:91:4d:f0::')
'Missing MAC Charcaters'
```

BobbyGR:
Saw an open issue. wasn't sure if it was still needed.
when I gave helper.

from napalm_base import helpers
print helpers.mac('1234.1234.12ab')

would just give me  u='', figured it was still an issue.  

I've never seen the mac missing characters like the one you are saying.  I can remove the check and just add the logic, it would give any numbers give the format.

mirceaulinic:
No worries @BobbyGR thanks for trying this out - this is how you can learn :)
Meanwhile, I had a closer look and some slight changes had to be applied though: https://github.com/napalm-automation/napalm-base/pull/91

PR #91, opened 2016-09-17
Enhance helpers

mirceaulinic:
@dbarrosop @ktbyers please review

### DIFF
--- a/napalm_base/helpers.py
+++ b/napalm_base/helpers.py
@@ -11,7 +11,6 @@ import textfsm
 from netaddr import EUI
 from netaddr import mac_unix
 from netaddr import IPAddress
-from netaddr.core import AddrFormatError
 
 # local modules
 import napalm_base.exceptions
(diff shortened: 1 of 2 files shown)


inetuid:
What would be involved in changing mac() and ip() to return a class so that they could be used in the test models? At present they return unicode so it's difficult to tell if a driver used them or not.

mirceaulinic:
AFAIK there's no driver using them yet. Their only purpose is to return a standard representation across all drivers as currently they depend on the vendor.

inetuid:
I don't think I was clear on my intent. Currently napalm_base/test/models.py has something like:

```
arp_table = {
    'interface' : unicode,
    'mac'       : unicode,
    'ip'        : unicode,
    'age'       : float
}
```

I was wondering whether it could be changed to something like:

```
arp_table = {
    'interface' : unicode,
    'mac'       : mac_class,
    'ip'        : ip_class,
    'age'       : float
}
```

At present I don't think this is possible as mac() and ip() return unicode() so would pass current tests.

mirceaulinic:
I see what you mean and does make sense. But if we are doing this right now, all the tests will fail :) The only way we can avoid this to happen would be to synchronise and update all of them at once.

dbarrosop:
I think what you are talking would be easier to do if we would have on out tests something like

```
assert is_valid_mac(result['mac_address'])
```

We can probably figure something out for the future but I would leave existing methods as they are.

PR #92, opened 2016-09-17
Fix for bug when not installed.

inetuid:
As per slack conversation during napalm-hackathon-2016.

### DIFF
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -19,14 +19,17 @@ import sys
 import inspect
 import importlib
 
-from pkg_resources import get_distribution
+import pkg_resources
 
 # NAPALM base
 
 from napalm_base.base import NetworkDriver
 from napalm_base.exceptions import ModuleImportError
 
-__version__ = get_distribution('napalm-base').version
+try:
+    __version__ = pkg_resources.get_distribution('napalm-base').version
+except pkg_resources.DistributionNotFound:
+    __version__ = "Not installed"
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


mirceaulinic:
Hi @inetuid can you expand more on this issue? Why would you find useful to return ```__version__ = "Not installed"```?

The point of ```__version__``` is to be able to do ```napalm_base.__version__```. But if uninstalled you can't use it anyway:

```python
>>> import napalm_base
ImportError: No module named napalm_base
>>> napalm_base.__version__
NameError: name 'napalm_base' is not defined
```

Unless you have something different in mind.

inetuid:
It's for when it's not installed via pip but you have a version (from git clone for example) in PYTHONPATH. It fails as (AIUI) pkg_resources.get_distribution('napalm-base') can't find the non-pip-installed version.

mirceaulinic:
> It's for when it's not installed via pip but you have a version (from git clone for example)

In that case it means you have already napalm-base installed, while ```__version__ = "Not installed"``` would suggest it is not installed.

However, I am running a custom version of napalm-base that does not have any of the recent changes:

```python
>>> import napalm_base
>>> napalm_base.__version__
AttributeError: 'module' object has no attribute '__version__'
```

And extracting the version info does work:

```python
>>> from pkg_resources import get_distribution
>>> get_distribution('napalm-base').version
'0.15.0cf'
```

I tried and it works installing using ```python setup.py install``` and ```pip install -e .```. Do you have a different use case I miss?

PR #93, opened 2016-09-17
Add file_copy method

itdependsnetworks:
Add Method that to file_copy any file to a network device. This port's over functionality from pyntc to napalm.

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -73,6 +73,20 @@ class NetworkDriver(object):
             print(epilog)
         raise exc_type, exc_value, exc_traceback
 
+
+    def file_copy(self, src, dest=None, **kwargs):
+        """Send a local file to the device.
+        Args:
+            src (str): Path to the local file to send.
+        Keyword Args:
+            dest (str): The destination file path to be saved on remote flash.
+                If none is supplied, the implementing class should use the basename
+                of the source path.
+            file_system (str): Supported only for IOS and NXOS. The file system for the
+                remote fle. Defaults to 'flash:' for IOS and 'bootflash:' for NXOS.
+        """
+        raise NotImplementedError
+
     def open(self):
         """
         Opens a connection to the device.


napalm_base/base.py
```diff
@@ -73,6 +73,20 @@ def __raise_clean_exception(exc_type, exc_value, exc_traceback):
...
+
+    def file_copy(self, src, dest=None, **kwargs):
+        """Send a local file to the device.
+        Args:
+            src (str): Path to the local file to send.
+        Keyword Args:
+            dest (str): The destination file path to be saved on remote flash.
+                If none is supplied, the implementing class should use the basename
+                of the source path.
+            file_system (str): Supported only for IOS and NXOS. The file system for the
```
dbarrosop: Can we piggyback on the optional argument passed upon creation or even better, encode it on the destination? For example:

    file_copy(src="source/to/file", dest="flash0://file")

itdependsnetworks: That works, I'll update it.

napalm_base/base.py
```diff
@@ -73,6 +73,20 @@ def __raise_clean_exception(exc_type, exc_value, exc_traceback):
             print(epilog)
         raise exc_type, exc_value, exc_traceback
 
+
+    def file_copy(self, src, dest=None, **kwargs):
+        """Send a local file to the device.
```
mirceaulinic: Why would you limit only to sending files to the device? I find myself many times copying files from the device. I think we should make this method bidirectional.

itdependsnetworks: ok, if I do that, then perhaps it should be:
def file_copy(self, local_file, remote_file=None, direction='push' **kwargs):

What do you think?

dbarrosop:
As @ktbyers mentioned, it might make sense to have instead `file_put` and `file_get` operations. It would simplify the API and the code. What do you think? That would also allow us to easily add `file_delete` and maybe other `file` operations if we decide to do so. It also might make sense to have those operations in `napalm-base` rather than in the drivers?

mirceaulinic:
> As @ktbyers mentioned, it might make sense to have instead file_put and file_get operations. It would simplify the API and the code

Makes sense to me.

> It also might make sense to have those operations in napalm-base rather than in the drivers?

As all are going to be SCP based, they can be defined under helpers and then just called in the NetworkDriver base class, like we are doing for ```load_template```.
We only need to avoid vendor-specific details, e.g. things like ```flash://``` etc.

mirceaulinic:
```file_delete``` might be more vendor specific as for JunOS, for example you can call ```file delete <filename>``` through netconf. But we can leave this aside for the moment.

**Comment by ktbyers** [comment]
Yes, I think we might want to plan this a bit more. For example, do we want this to be all SCP based or SCP-based by default (in base) with possibility of using API instead (if possible) in the vendor specific driver.

**Comment by itdependsnetworks** [comment]
If agreed, I'll start with the file_put method. 

`def file_put(self, local_file, remote_file=None, **kwargs):`

I'll start with the 4 ported from pyntc. 
The similarly for file_get: 

`def file_put(self, local_file, remote_file=None, **kwargs):`

I'll ping @jedelman8 for his input, rather then just turning around and tapping on shoulder :)

**Comment by mirceaulinic** [comment]
I'm fine with SCP - I think all vendors should support at least this for file transfer. I know for sure that it works fine for juniper, iosxr and nxos... not for pluribus but this one is a special case that we can ignore.
One extra feature we can have under each driver would be to determine the default storage directory, e.g.: ```/var/home/<username>/``` on junos, ```disk0a:/usr``` on iosxr etc. and if the argument ```remote_file``` is not specified will copy under the default dir with the original name if possible. Does this make any sense?

**Comment by itdependsnetworks** [comment]
Yes that does make sense. My thought process is it **wouldn't** be: 

`def file_put(self, local_file, remote_file='flash0:', **kwargs):`

Instead it would check if there is a colon, and then split out the file system from file name. Otherwise just use that as destination file. So if you want to send nyc-rs01.conf and push it as `flash:\\lastknowngood.conf` you would just send as `remote_file=lastknowngood.conf`. However if you wanted to send to startup_config then it would be `nvram:\\startup-config`

**Comment by ktbyers** [comment]
A couple of other advantages of using Netmiko, you will get SSH key support and SSH config file support as part of this (at least from the hooks into Paramiko). Many of these items are not easy to solve.

**Comment by mirceaulinic** [comment]
I have mixed feelings about using netmiko features in this case. Although very tempting to use something already done (as it's much easier), I'm not sure we really want to have netmiko as a requirement in napalm-base.
Also, the methods under the FileTransfer class do not seem to be generic enough, but they seem to be bespoken and constrained to vendors like Cisco etc. - for example the method ```remote_file_size``` which is called in the ```__init__``` would probably fail when run against a Junos device as the comments suggest operations such ```# Strip out "Directory of flash:/filename line``` or ```# Format will be 26  -rw-   6738  Jul 30 2016 19:49:50 -07:00  filename```.
On the other hand, from the [scp library](https://pypi.python.org/pypi/scp) we can simply call ```put``` and ```get``` especially given that we can use the underneath paramiko connection already open (in drivers such _napalm-junos_ (used to open the netconf session), _napalm-iosxr_, _napalm-ios_ etc.).

**Comment by ktbyers** [comment]
@mirceaulinic Yes, there is definitely work I have to do to make Netmiko's scp_handler more generic. I don't think that will be too hard, however (and I am probably going to do this anyways).

**Comment by mirceaulinic** [comment]
As previously said, we can re-use the already established connection, don't see any gain by doing things overly complicated:

```python
>>> from napalm_base import get_network_driver
>>> d = get_network_driver('junos')
>>> e = d('127.0.0.1', 'napalm', 'Napalm123', optional_args={'port': 12202, 'config_lock': False})
>>> e.open()
>>> e.device._conn._session.transport
<paramiko.Transport at 0x8f00f90L (cipher aes128-ctr, 128 bits) (active; 1 open channel(s))>
>>> from scp import SCPClient
>>> scp = SCPClient(e.device._conn._session.transport)
>>> scp
<scp.SCPClient object at 0x1096a9e90>
>>> scp.get('/var/tmp/example.log')
>>>
```

done.

dbarrosop:
If something is to be added to `napalm-base` we want whatever it introduces the least amount of dependencies. We don't want to bloat the requirements. I like @mirceaulinic example, although it would still add three dependencies for a couple of methods : (

I wonder if it's worth having this as part of `napalm-utils` or something like that. Something where we can do things like:

```
with open(driver(..)) as device:
    napalm_utils.file_put(device, local_file, remote_file)
```

mirceaulinic:
I am also very reticent about having many dependencies - we must bear in mind that napalm-base requires jinja2, gtestfxm and netaddr which is already too much. The solution I proposed would add would add one more: scp - even so I would prefer to keep away.

I have in mind the following solution:

Under ```napalm_base/helpers.py```:

```python
def file_get(transport, filepath):
    scp = SCPClient(transport)
    scp.get(filepath)
```

And the driver would call like, for example ```napalm_junos/junos.py```:

```python
napalm_base.helpers.file_get(self.device._conn._session.transport, '/path/to/file')
```

Does this make sense?

mirceaulinic:
I am into the idea of a separate package for utilities and require this library only when needed. On the other hand, I think we already use our helpers everywhere, unless this would be the first exception.

ktbyers:
I like the idea of having a separate package napalm-utils and making the napalm-base very lightweight.

Note, on the scp dependencies. Paramiko and its requirement of cryptography are the big added dependencies (applies both to scp and Netmiko). Paramiko can cause install problems (at least historically when it was using pycrypto).

There is going to be quite a bit of glue code required to deal with device file systems, executing remote md5s, checking if file already exists, checking if space exists, etc. 

I seem to recall I ran into issues with trying to reuse the existing SSH connection. If I recall correctly, I had to abandon doing this when I was building the scp handler for Netmiko.

dbarrosop:
We can probably have most of the code in `utils` of `base` (depending on what we decide on that front) and make each driver implement some simple functions to help dealing with filesystems. Something like, `_resolve_filename(self, filename)` that helps populating some information for scp to work. You probably know better what the differences are so feel free to highlight which "vendor specific" code might be needed.

dbarrosop:
As apparently we have some consensus here I am going to create `napalm-utils` and ask @itdependsnetworks to move his PR there. Could any of you help him with the changes that will be needed, please?

@itdependsnetworks seems you opened a can of worms : ) Just kidding, we will ask you to move the PR to a different repo and do some changes. I think this makes more sense to have it under a 'utils' package than directly on every driver.

dbarrosop:
Repo is here https://github.com/napalm-automation/napalm-utils

@itdependsnetworks wait for instructions, any of us will provide with an example and guide you through the change.

Thanks for your patience : )

dbarrosop:
Ok, so I have been giving more thoughts to this and a couple of notes:

1. I am now with the idea that `file_put` and `file_get` should be part of the driver itself. It's easier to use it than have it as a helper.
2. We can make `paramiko` import optional on `napalm-base`. This is to avoid adding huge dependencies to all the drivers. We can do:

```
try:
   has_paramiko = False
except ImportError:
   paramiko = None
```

And then in the code:

```
def file_put(...):
    if not paramiko:
        raise SomeNewException("Please, install paramiko if you want to be able to upload/download files")
    ...
```

3. I haven't seen any device that let's you upload devices via HTTP so I think it's safe to assume scp/sftp is the only thing we need to support. In any case, if we ever want to support something else  and is generic enough we can add it and use an optional arg or just overload the method if it's not generic enough.

Thoughts?

ogenstad:
>I haven't seen any device that let's you upload devices via HTTP so I think it's safe to assume scp/sftp is the only thing we need to support. In any case, if we ever want to support something else and is generic enough we can add it and use an optional arg or just overload the method if it's not generic enough.

With this I'm able to upload a file to a Cisco ASA:
```
curl -k -u patrick --request POST --data-binary "@testfile.txt" https://172.29.52.33/admin/disk0/testfile.txt
```
Having said that I've never used it. However it's also possible to download files from the ASA `file_get`, specifically with capture files you can download them the device using https. With the https option you can also add /pcap to the path to download the files in pcap format instead of the raw format. This isn't possible using scp.

Also, there's nothing to say that http won't be used more in the future. 

Another thing is that copying files using paramiko is slow, it could make sense to have it idempotent so that the `file_put` only copies the file if the exact same file isn't already on the device. In order to have checks for this it might be hard to this as a generic top level function.

dbarrosop:
> With this I'm able to upload a file to a Cisco ASA

Fair enough, the operations can support a "transport" parameter to specify HTTP vs SCP. Both should be generic enough to work out of the box. We will need kwargs here to be able to pass things to the underlying request library. However, I suggest here start with the `scp` and then iterate.

> Another thing is that copying files using paramiko is slow, it could make sense to have it idempotent so that the file_put only copies the file if the exact same file isn't already on the device

We can have a `get_md5` method to do that although that's gonna be platform independent. We can use it if available  for idempotency and where it's not available just push the file directly.

itdependsnetworks:
The md5 check will be done for sure, the only question last time was whether or not that was a method available via napalm, or done in the background.

dbarrosop:
Yeah, I am starting to think it makes sense to have it here.

ktbyers:
I think we should have MD5 comparison (local and remote), file size verification (is there enough space on the device), maybe capability to automatically enable/disable SCP.

itdependsnetworks:
Sounds like a plan. Just wasn't clear, where this goes? in utils? if so how is that structured?

dbarrosop:
I think we should define a set of methods as @ktbyers suggested in napalm-base and then just override them on each driver. We could potentially have a base implementation here as most of the time file management is going to be done using scp regardless of the driver.

ktbyers:
@dbarrosop Yes, I agree on that...I also think for device types that are using Netmiko that we should use the underlying Netmiko scp_handler (so we don't have to reinvent the wheel). This will also get us the SSH control channel using Netmiko.

For platforms using APIs, we will have to figure out what the best method to implement and how much code we can re-use from Netmiko and/or NTC file copy.

dbarrosop:
Do you mind project-managing this one? I think you have a way better understanding on how to best fix this than me :) We can create a feature-branch like we are doing with the yang stuff and work in there before merging into develop.

dbarrosop:
also, create a label and tag all the related issues and PRs with it so we can easily track everything down with zenhub.

Thanks!

ktbyers:
Okay, have assigned @itdependsnetworks and myself to this issue.

ktbyers:
Moving to separate issue to track this.

PR #98, opened 2016-09-18
All the code should pass pylama tests

dbarrosop:


ktbyers:
@dbarrosop I think we should make the skip line be:

```
[pylama]
linters = mccabe,pep8,pyflakes
ignore = D203,C901
skip = */openconfig_bindings/*

[pylama:pep8]
max_line_length = 100
```

So it won't try to pylama the openconfig_bindings in ./build

ktbyers:
Everything else is fine besides one recommended change to setup.cfg.

dbarrosop:
Fixed it. I will be removing openconfig soon, probably this week but I figured it was worth merging this in the meantime.

PR #104, opened 2016-09-28
Add python3 support to develop branch

ktbyers:
Feel free to review / test

Don't merge until discussing with me. I want to do some additional review.


dbarrosop:
Haven't reviewed yet but this is huge!!! You have no idea how happy this makes me : )

.travis.yml
```diff
@@ -2,11 +2,14 @@ language: python
...
   - 2.7
+  - 3.4 
 
 install:
   - pip install .
   - pip install -r requirements-dev.txt
-  - pip install -r test/unit/requirements.txt
+  - pip install -r requirements.txt
+  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then pip install -r test/unit/requirements.txt; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 3.4 ]]; then pip install -r test/unit/requirements_py3.txt; fi
```
dbarrosop: Why is this necessary?

napalm_base/__init__.py
```diff
@@ -37,9 +44,9 @@
     'NetworkDriver'  # also export the base class
 ]
 
-
 if not(sys.version_info.major == 2 and sys.version_info.minor == 7):
```
dbarrosop: It's probably clearer as:

```
if not(sys.version_info.major == 2 and sys.version_info.minor == 7) and not(sys.version_info.major == 3):
    raise RuntimeError('NAPALM requires Python 2.7 or Python3')
```

Which by the way, I realized Python <2.7 doesn't support that code anyway:

```
>>> if not(sys.version_info.major == 2 and sys.version_info.minor == 7) and not(sys.version_info.major == 3):
...      raise RuntimeError('NAPALM requires Python 2.7 or Python3')
...
AttributeError: 'tuple' object has no attribute 'major'
```

napalm_base/helpers.py
```diff
@@ -104,6 +112,7 @@ def textfsm_extractor(cls, template_name, raw_text):
     )
 
     try:
+        print(template_path)
```
dbarrosop: I guess this is a leftover

napalm_base/helpers.py
```diff
@@ -157,8 +163,7 @@ def find_txt(xml_tree, path, default=''):
                 value = xpath_result
     except Exception:  # in case of any exception, returns default
         value = default
-
-    return unicode(value)
```
dbarrosop: this is gonna hurt. We should probably change it for `u'{}'.format(value)`. A lot of code (specially tests) assumes unicode. Have you tested running `napalm-ios` unittests with this version of `napalm-base`

napalm_base/test/base.py
```diff
@@ -12,15 +12,22 @@
...
+from napalm_base.test import models
 from unittest import SkipTest
 
+# For unit tests only!
+if sys.version[0] == '3':
+    unicode = str
+
 
-class TestConfigNetworkDriver:
+class TestConfigNetworkDriver(object):
```
dbarrosop: Why did you change this to the old python class style?

napalm_base/test/getters.py
```diff
@@ -1,20 +1,22 @@
...
 import helpers
-
 import models
-
-
 import pytest
 
+# For unit tests only!
+if sys.version[0] == '3':
+    unicode = str
```
dbarrosop: hahaha! This is what I feared a comment a couple of comments ago : )

.travis.yml
ktbyers: @dbarrosop napalm-base tries to load all of the napalm drivers as part of its testing.

```
$ cat test/unit/requirements.txt 
ddt
napalm-eos
napalm-fortios
napalm-ibm
napalm-ios
napalm-iosxr
napalm-junos
napalm-nxos
napalm-pluribus
```

This causes the Travis-CI test to fail under Python3 (as those drivers don't support Python3 yet i.e. we are going to add Python3 support to those drivers one-by-one).

There is also this corresponding change that I made to TestGetNetworkDriver:
```
class TestGetNetworkDriver(unittest.TestCase):
    """Test the method get_network_driver."""

    if sys.version[0] == '2':
        # All drivers support python2
        network_drivers = ('eos', 'eos', 'fortios', 'ibm', 'ios', 'iosxr', 'IOS-XR', 'junos',
                           'nxos', 'pluribus', 'panos')
    elif sys.version[0] == '3':
        # Drivers that support python2 and python3
        network_drivers = ()
```

napalm_base/__init__.py
ktbyers: Okay, fixed...included causing AttributeError in this context to raise the same RuntimeError.

napalm_base/helpers.py
ktbyers: Removed

napalm_base/test/base.py
ktbyers: I didn't...I changed it to use new-style classes.

This in Python2 is an old-style class:
```
-class TestConfigNetworkDriver:
```

This is new-style class:
```
+class TestConfigNetworkDriver(object):
```

Python3 confuses the matter as they only have new-style classes and both of the above are new-style classes.

napalm_base/helpers.py
ktbyers: Yes the:

```
from __future__ import unicode_literals
```

Makes all string literals to be unicode. So in Python2 anything like:

```
my_str = 'whatever'
```

Would automatically be a unicode and no need to do my_str = u'whatever'. See comment below about test code and unicode.

napalm_base/test/getters.py
ktbyers: Yes, I didn't want to re-write all the test code that checks for type unicode as there is a lot of it.

I didn't really love this, but it seemed like the least bad thing to do. I was okay with this in the test code, but wouldn't have been okay with it in non-test code.

Definitely open to discussions on it, however, as it is a kludge.

napalm_base/__init__.py
```diff
@@ -14,15 +14,29 @@
 
 """napalm_base package."""
 
+# Python3 support
+from __future__ import print_function
+from __future__ import unicode_literals
+try:
```
mirceaulinic: Those try-catch blocks appear here in a couple of places and they will be needed in all napalm drivers for sure.

I would propose the following: why not define them in a napalm-base module? For example under ```napalm_base/utils```. We can call it ```six.py``` perhaps.

And there we can use the following:

```python
if sys.version_info.major == 3:
    basestring = str
    etc.
```

Or even we can add one extra layer of abstraction:

```python
if sys.version_info.major == 3:
    string_type = str
else:
    string_type = basestring
```

Similarly with other datatypes.

Additionally we can have also some useful constants that can be reused to test the Python version and feels somehow more natural:

```python
PY2 = sys.version_info.major == 2
PY3 = sys.version_info.major == 3
```

What do you think?

mirceaulinic: Same with unicode and we could call it:

```python
if sys.version_info.major == 3:
    string_type = str
    txt_type = str
else:
    string_type = basestring
    txt_type = unicode
```

.travis.yml
dbarrosop: Ok, makes sense.

napalm_base/__init__.py
dbarrosop: awesome : )

napalm_base/helpers.py
dbarrosop: ok : )

napalm_base/test/base.py
dbarrosop: Mmm, not sure I understand:

https://docs.python.org/3/tutorial/classes.html#class-definition-syntax

According to the documentation the syntax for defining a class is:

```
class ClassName:
   ...
```

napalm_base/test/getters.py
dbarrosop: it's alright, one step at a time. Create an issue to track it and we will take a look once this is merged.

napalm_base/__init__.py
dbarrosop: Would that allow us to import just one thing and get all the compatibility requirements we have solved?

mirceaulinic: I think so - instead of repeating the try-expect section and testing against the python version in all modules, we could simply import and use ```napalm_base.utils.six.str_type``` for example.

napalm_base/test/base.py
ktbyers: Yes, you linked to the Python3 docs. The Python2 docs are also confusing, however, which is not helpful.

Here are a couple of references:
https://wiki.python.org/moin/NewClassVsClassicClass
http://stackoverflow.com/questions/54867/what-is-the-difference-between-old-style-and-new-style-classes-in-python

So Python2, new-style classes are created in this form:

```
# Must inherit from object (or from other class which ultimately inherits from object)
class MyClass(object):
```

In Python3, all classes are new-style classes, there are no old-style classes. So these two are equivalent:

```
# New style in Python3
class MyClass:
```

Also new-style in Python3
```
class MyClass(object):
```

In order to write compatible code, we should just write it in the Python2 style i.e. inherit from 'object'.

napalm_base/__init__.py
ktbyers: Let me look at this. I like the following two items:

```
PY2 = sys.version_info.major == 2
PY3 = sys.version_info.major == 3
```

and

```
if sys.version_info.major == 3:  # or if PY3:
    string_type = str
    txt_type = str
else:
    string_type = basestring
    txt_type = unicode
```

ktbyers: Okay updated to use the following:

./napalm_base/utils/py23_compat.py

With the following code (this made it more consistent with six):

```
PY2 = sys.version_info.major == 2
PY3 = sys.version_info.major == 3

if sys.version_info.major == 3:
    string_types = (str,)
    text_type = str
else:
    string_types = (basestring,)  # noqa
    text_type = unicode  # noqa
```

Didn't want to call the module 'six' as that would cause confusion as to whether we were using the six library or not.

ktbyers:
I am also regression testing the following:

1. Cisco IOS driver using PY2 (both getter and config)
2. Cisco IOS driver using PY3 (both getter and config)

ktbyers:
Okay, I think this is ready to go once I complete the above tests.

ktbyers:
Okay, tests passed. This PR should be ready to go from my perspective. 

Could either of you regression test it against EOS, IOS-XR, Juniper--one or two of these at least the getters?

mirceaulinic:
Sure @ktbyers I am going to make some tests tomorrow!

ktbyers:
@mirceaulinic Thanks.

napalm_base/__init__.py
```diff
@@ -14,17 +14,28 @@
...
+except AttributeError:
+    raise RuntimeError('NAPALM requires Python 2.7 or Python3')
 
+# NAPALM base
 from napalm_base.base import NetworkDriver
 from napalm_base.exceptions import ModuleImportError
+from napalm_base.utils import py23_compat
 
 try:
     __version__ = pkg_resources.get_distribution('napalm-ios').version
```
mirceaulinic: I guess here should be:

```python
__version__ = pkg_resources.get_distribution('napalm-base').version
```

napalm_base/base.py
```diff
@@ -12,6 +12,10 @@
 # License for the specific language governing permissions and limitations under
 # the License.
 
+# Python3 support
+from __future__ import print_function
+from __future__ import unicode_literals
```
mirceaulinic: Do we need to import both?

napalm_base/__init__.py
ktbyers: Strange...I will fix this. I swore I fixed that previously.



Kirk

napalm_base/base.py
ktbyers: I think we should have both:
1. The first one to ensure that any print statement errors are quickly caught.
2. The second one to ensure that all string literals are unicode.

Of the two the unicode_literals one is more important.

mirceaulinic: I understand, I am not commenting against them - I actually encourage. I am just thinking if would be possible to find a way to avoid importing them everywhere... something similar to the py23_compat

napalm_base/__init__.py
ktbyers: Okay, fixed.

mirceaulinic:
I tested it against napalm-iosxr 0.3.0 and I confirm it works almost fine:

```python
>>> k = e.get_arp_table()[0]
>>> k
{'interface': u'BVI10', 'ip': '1.2.3.4', 'mac': EUI('E0:AC:F1:67:48:D8'), 'age': 0.0}
>>> k['mac']
EUI('E0:AC:F1:67:48:D8')
>>> type(k['mac'])
<class 'netaddr.eui.EUI'>
```

Whilst the mac address should be represented as string instead of EUI object. Same for ```get_mac_address_table```.
So there's a change needed to be applied in the mac helper.


I would also like to test against EOS as there are some textfsm templates needed in the getters.

mirceaulinic:
Tested also against napalm-eos 0.3.0 and napalm-junos and I confirm it works fine.

ktbyers:
@mirceaulinic Is this an issue related to my changes or some other bug? Not clear to me.

```
Whilst the mac address should be represented as string instead of EUI object. Same for get_mac_address_table.
So there's a change needed to be applied in the mac helper.
```

mirceaulinic:
It is related to this changes @ktbyers.

Using napalm-base 0.17.0 and napalm-iosxr 0.3.0:

```python
>>> e.get_arp_table()[0]['mac']
u'E0:AC:F1:67:48:D8'
```

napalm_base/helpers.py
```diff
@@ -214,7 +215,7 @@ def mac(raw):
             flat_raw=flat_raw,
             zeros_stuffed='0'*(12-len(flat_raw))
         )
-    return unicode(EUI(raw, dialect=_MACFormat))
+    return EUI(raw, dialect=_MACFormat)
```
mirceaulinic: Here's what did not need to be changed - should return ```unicode``` instead of ```EUI``` object, or better the ```py23_compat.text_type```

napalm_base/helpers.py
```diff
@@ -237,5 +238,4 @@ def ip(addr):
         >>> ip('2001:0dB8:85a3:0000:0000:8A2e:0370:7334')
         u'2001:db8:85a3::8a2e:370:7334'
     """
-
-    return unicode(IPAddress(addr))
+    return str(IPAddress(addr))
```
mirceaulinic: Same here - let's make sure we use the py23_compat datatypes

ktbyers: Okay, let me fix.

ktbyers:
@mirceaulinic Okay, I think I fixed those type issues that I introduced earlier.

mirceaulinic:
@ktbyers I tested the napalm-base and looks good, except:

```
$ nosetests --with-coverage --cover-package napalm_junos -v TestJunOSDriver:TestGetterJunOSDriver.test_get_route_to
test_get_route_to (TestJunOSDriver.TestGetterJunOSDriver) ... ERROR

======================================================================
ERROR: test_get_route_to (TestJunOSDriver.TestGetterJunOSDriver)
----------------------------------------------------------------------
  File "/state/home/mircea/kirk/napalm-base/napalm_base/test/base.py", line 388, in test_get_route_to
    get_route_to = self.device.get_route_to(destination=destination, protocol=protocol)
  File "/state/home/mircea/napalm-test/local/lib/python2.7/site-packages/napalm_junos/junos.py", line 923, in get_route_to
    raise TypeError('Please specify a valid destination!')
TypeError: Please specify a valid destination!

Name                                Stmts   Miss  Cover
-------------------------------------------------------
napalm_junos.py                         2      0   100%
napalm_junos/junos.py                 663    139    79%
napalm_junos/utils.py                   0      0   100%
napalm_junos/utils/junos_views.py       4      0   100%
-------------------------------------------------------
TOTAL                                 669    139    79%
----------------------------------------------------------------------
Ran 1 test in 0.002s
```

Which fails because the method ```get_route_to``` expects ```destination``` as str: https://github.com/napalm-automation/napalm-junos/blob/develop/napalm_junos/junos.py#L960 I will change this in napalm-junos 0.4.0 and napalm-iosxr 0.3.1 (will release today).

There are some other failures:

- napalm-iosxr:

```
$ nosetests --with-coverage --cover-package napalm_iosxr -v TestIOSXRDriver:TestGetterIOSXRDriver.test_get_users
test_get_users (TestIOSXRDriver.TestGetterIOSXRDriver) ... FAIL

======================================================================
FAIL: test_get_users (TestIOSXRDriver.TestGetterIOSXRDriver)
----------------------------------------------------------------------
  File "/state/home/mircea/kirk/napalm-base/napalm_base/test/base.py", line 486, in test_get_users
    self.assertTrue(result)
AssertionError: False is not true
-------------------- >> begin captured stdout << ---------------------
key: password
model_class: <type 'unicode'>
data_class: <type 'str'>

$ nosetests --with-coverage --cover-package napalm_iosxr -v TestIOSXRDriver:TestGetterIOSXRDriver.test_get_config
Test get_config method. ... FAIL

======================================================================
FAIL: Test get_config method.
----------------------------------------------------------------------
  File "/state/home/mircea/kirk/napalm-base/napalm_base/test/base.py", line 520, in test_get_config
    assert self._test_model(models.config, get_config)
AssertionError:
-------------------- >> begin captured stdout << ---------------------
key: running
model_class: <type 'unicode'>
data_class: <type 'str'>
key: startup
model_class: <type 'unicode'>
data_class: <type 'str'>
key: candidate
model_class: <type 'unicode'>
data_class: <type 'str'>

--------------------- >> end captured stdout << ----------------------
```

- napalm-eos:

```
test/unit/test_getters.py::TestGetter::test_get_users[normal] <- ../kirk/napalm-base/napalm_base/test/getters.py key: password
model_class: <type 'unicode'>
data_class: <type 'str'>
FAILED
test/unit/test_getters.py::TestGetter::test_get_config[normal] <- ../kirk/napalm-base/napalm_base/test/getters.py key: running
model_class: <type 'unicode'>
data_class: <type 'str'>
key: startup
model_class: <type 'unicode'>
data_class: <type 'str'>
key: candidate
model_class: <type 'unicode'>
data_class: <type 'str'>
FAILED
```

It looks like you changed the datatype expected for:

- password in ```get_users```
- running, candidate, startup for ```get_config```

I think we should be consistent and use ```text_type``` everywhere. That means adjusting all drivers.
With thee said, I think we can accept this changes today and next releases to reflect these datatypes, otherwise the tests will fail after this branch is merged.

napalm_base/test/base.py
dbarrosop: Oh god, that's stupid! I had gotten this wrong all this time... Why would someone adopt a new style of something and then build a new version of the programming language and adopt the old style? :S

napalm_base/__init__.py
dbarrosop: That's great, we should document thoroughly how to use it and when. I don't want that library to become like other utils, that people doesn't know it exists : (

napalm_base/base.py
dbarrosop: AFAIK you can't just import in one place, you have to do it per module as it modifies the namespace of the module itself.

napalm_base/utils/py23_compat.py
```diff
@@ -0,0 +1,15 @@
+"""Simplify Python3 compatibility. Modeled after six behavior for small set of things"""
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import sys
+
+PY2 = sys.version_info.major == 2
+PY3 = sys.version_info.major == 3
+
+if sys.version_info.major == 3:
```
mirceaulinic: We can use ```if PY3:``` for simplicity

napalm_base/utils/py23_compat.py
```diff
@@ -0,0 +1,15 @@
...
+    string_types = (str,)
```
mirceaulinic: Not sure why do you want this to be a tuple? 
I would have defined as: ```string_type = str``` for py3 and ```string_type = basestring``` for py2

dbarrosop:
Are we going to break something if we merge this?

mirceaulinic:
Yes @dbarrosop the tests will fail for all drivers implementing ```get_users``` and ```get_config``` because they are using the develop branch of the napalm-base, e.g.: https://github.com/napalm-automation/napalm-eos/blob/develop/.travis.yml#L7

But this can be a motivation to adjust in order to be  consistent :)

dbarrosop:
Before merging this one, can we put on an issue what has to be done? I can "spread" the issue to repos then for traceability. I guess we have to do, at least:

1. Enable tests for python3.
2. Do whatever is necessary to fix tests for both py2 and py3?
3. Document py23_compat module (this is not driver specific but should deserve a dedicated issue)

Something else I am missing?

mirceaulinic:
Sure, sounds good to me.
Can you please expand more on "enable tests for python3"? I am not sure what would I have to do?

napalm_base/utils/py23_compat.py
ktbyers: @mirceaulinic I looked at what they did in six (and thought it made sense to pattern on what they did). Behavior is the same when used in isinstance.

ktbyers:
The issue is we still have some things in Python2 that were returning strings and not unicode and we are type checking them (and now expecting them to be unicode).

We just need to make the change in the Python2 underlying driver to make them return unicode and not str. I will fix the ones you reference above. 

There could be others in other drivers.


@mirceaulinic I am not sure I understand this?

```
I think we should be consistent and use text_type everywhere.
```

All strings will now be unicode in both Python2 and Python3 (Python3 calls unicode 'str'). We should avoid type casting strings unless forced into it. Agreed, however, if we must type cast something, then we need to use 'text_type'.

napalm_base/utils/py23_compat.py
mirceaulinic: Alright, cool :)

ktbyers:
FYI, I will regression test all of the drivers on the Mock test and stage any needed fixes. So at least Travis will pass.

This does illustrate a testing problem, however (i.e. change napalm-base and break things in drivers that doesn't get caught in the napalm-base test).

mirceaulinic:
> This does illustrate a testing problem, however (i.e. change napalm-base and break things in drivers that doesn't get caught in the napalm-base test).

Sounds like sub-optimal and sort of funny. But we knew from the beginning when we split napalm about this egg and chicken type problem and unfortunately we need to accept a tradeoff.

I identify the possibilities:

- dependency on develop branch of napalm-base: can easily introduce new methods and test on the fly - and eventually adjust the model as needed. But when something breaks the develop napalm-base will propagate the failure in the driver tests.

- dependency on the latest release of napalm-base: whenever a new method is introduced, a new napalm-base version is released (which sometimes might be a bit painful as the models might need adjustments and will end up having a couple of releases of the napalm-base just to introduce a new method). But has the big advantage that the tests of the drivers will use only a stable version of the napalm-base.

I would personally go for the second. What do you think?

dbarrosop:
> Can you please expand more on "enable tests for python3"? I am not sure what would I have to do?

The changes needed in travis to ensure we test both py2 and py3

> Sounds like sub-optimal and sort of funny. But we knew from the beginning when we split napalm about this egg and chicken type problem and unfortunately we need to accept a tradeoff.

Only the testing is going to break due to the differences of the string types, nothing else should break so this is mostly an issue for us that shouldn't affect users.

> I would personally go for the second. What do you think?

I'd go actually with a variant of the second, we should probably pin `napalm-base` on each driver to make sure a change on `napalm-base` doesn't break a yet-to-update driver. Thoughts?

ktbyers:
@dbarrosop Yes, I like the idea of pinning the napalm-base driver version so that this update in napalm-base doesn't break the other drivers. We can then correct this as we add support to Python3 for each of the drivers (or if we run into a driver that can't support Python3 right now at least make sure Python2 tests properly with new napalm-base).

mirceaulinic:
Me too, I'm into this idea.

> The changes needed in travis to ensure we test both py2 and py3

Still not clear to me what.

I see @ktbyers added the following:

```
+  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then pip install -r test/unit/requirements.txt; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 3.4 ]]; then pip install -r test/unit/requirements_py3.txt; fi
```

But I assume we should have the same set of requirements for the tests, regardless of the python version. Is there anything I'm missing?

ktbyers:
@mirceaulinic

The issue is this file:

napalm-base/test/unit/requirements.txt. 

```
ddt
napalm-eos
napalm-fortios
napalm-ibm
napalm-ios
napalm-iosxr
napalm-junos
napalm-nxos
napalm-pluribus
napalm-panos
```

The travis-CI test in Python2 will try to install each one of these packages. We can't do this for Python3 since trying to install those packages on Python3 will fail. Those packages need to be installed because for Python2, the napalm-base GetNetworkDriver test loads each drivers.

So the original test would have a circular dependency for Python3--i.e. upgrading napalm-base to Python3 would require all of the sub-drivers to also work for Python3 (at the same time).

So in the interim, I want to decouple this. I want napalm-base to be upgraded to Python3. I then want to work on upgrading the individual drivers to Python3 across time. 

Eventually, when all of the drivers are upgraded to support both Python2 and Python3 we can eliminate the separate requirements file.

mirceaulinic:
I understand the reasoning for napalm-base, that's fine. 

My question is: do we need to do the same for the drivers? I don't see why, but perhaps I don't take some factors into consideration.
If not, what are the changes required for travis.yml? I still don't know :)

ktbyers:
@mirceaulinic Okay, misunderstood your question. Yes, the requirements file should be the same for drivers. I don't see any reason for them to be different.

dbarrosop:
You need to, at least, do this:

```
 python:
    - 2.7
 +  - 3.4 
```

So, let's do this, then, let's release a new napalm-$driver version where we "pin" napalm-base to `0.17.0` and then we go driver per driver make it py3 compatible : )

If you agree I will start creating issues on all drivers to track both pinning napalm-base and making the driver py3 compatible : )

ktbyers:
That sounds good to me...the pinning the napalm-base driver to 0.17.0 for both master and develop branch.

dbarrosop:
Done, when we are done with this one: napalm-base#107 we can merge this one too : )

dbarrosop:
I have also created issues everywhere to enable py3 support

dbarrosop:
Can't wait for this one...

dbarrosop:
When we merge this one we will have to go module by module testing it works and release a new version where requirements say it has to be

`napalm-base >= 0.18.0`

mirceaulinic:
Yes, let's do it this week as I have some more spare time to fix eventual issues (I expect napalm-junos to have a few and probably all drivers implementing ```get_users``` and ```get_config``` as the datatype was changed from ```str``` to ```unicode```).

dbarrosop:
Yeah, let's wait for ktbyers to be online. I want him to do the honors and merge this himself. I am saving some time this evening to look at `napalm-eos` and `napalm-fortios` ASAP and fix anything that might pop up : )

ktbyers:
@dbarrosop @mirceaulinic We all good to go on this?

dbarrosop:
Go go go!

mirceaulinic:
Yay!!! 🍰

PR #110, opened 2016-10-16
Dbarrosop/test framework revisit mocking

dbarrosop:
This PR has a major enhancement when it comes to patching the drivers for mock testing. Now you can create a FakeDriver that inherits from the real driver and just do the monkey patching yourself. Then you can tell the framework which objects were patched. Example:

```
class PatchedEOSDriver(eos.EOSDriver):
    """Patched EOS Driver."""

    def __init__(self, hostname, username, password, timeout=60, optional_args=None):
        super().__init__(hostname, username, password, timeout, optional_args)

        self.patched_attrs = ['device']
        self.device = FakeEOSDevice()


class FakeEOSDevice(BaseTestDouble):
    """EOS device test double."""

    def run_commands(self, command_list, encoding='json'):
        """Fake run_commands."""
        result = list()

        for command in command_list:
            filename = '{}.{}'.format(self.sanitize_text(command), encoding)
            full_path = self.find_file(filename)

            if encoding == 'json':
                result.append(self.read_json_file(full_path))
            else:
                result.append({'output': self.read_txt_file(full_path)})

        return result
```

and:

```
class PatchedROSDevice(ros.ROSDriver):
    """ROS device test double."""
    def __init__(self, hostname, username, password, timeout=60, optional_args=None):
        super().__init__(hostname, username, password, timeout, optional_args)

        self.patched_attrs = ['paramiko_transport', 'mikoshell']

    def open(self):
        self.paramiko_transport = FakeParamikoTransport()
        self.mikoshell = FakeMikoShell()
        self._datetime_offset = datetime.datetime.now() - datetime.datetime.now()


class FakeParamikoTransport(BaseTestDouble):
    def open_session(self):
        return FakeParamikoChannel()

    def close(self):
        pass

    def is_active(self):
        return True


class FakeMikoShell(BaseTestDouble):

    def command(self, command, *args, **kwargs):
        full_path = self.find_file(self.sanitize_text(command))
        return self.read_txt_file(full_path).splitlines()

    def exit(self, cmd):
        pass


class FakeParamikoChannel(BaseTestDouble):

    def close(self):
        pass

    def exec_command(self, command):
        self._exec_command = command

    def makefile(self, *args):
        return StringIO.StringIO(self.read_txt_file(self._exec_command))

    @staticmethod
    def set_combine_stderr(*args):
        pass

    @staticmethod
    def shutdown(*args):
        pass
```

This just gives more flexibility and control to the drivers.


dbarrosop:
@napalm-automation/council I need this reviewed

napalm_base/test/getters.py
```diff
@@ -71,9 +76,9 @@ def wrapper(cls, test_case):
         # This is an ugly, ugly, ugly hack because some python objects don't load
         # as expected. For example, dicts where integers are strings
```
mirceaulinic: ```integers are strings``` Whut? I guess you meant ```keys are integers```?

napalm_base/test/getters.py
```diff
@@ -52,8 +55,10 @@ def wrap_test_cases(func):
     """Wrap test cases."""
     @functools.wraps(func)
     def wrapper(cls, test_case):
-        cls.device.device.current_test = func.__name__
-        cls.device.device.current_test_case = test_case
+        for patched_attr in cls.device.patched_attrs:
```
mirceaulinic: Is ```cls.device.patched_attrs``` a list of all attributes of the method?

dbarrosop: yes, it's the list of patched attributes, in the case of eos it's just `device` but others might have more. For example, IBM had two.

napalm_base/test/getters.py
dbarrosop: Yes, your phrasing is correct. Problem was that python let you do this, `{ 1: 'asdads' }` (and we are doing it as a matter of fact) but json doesn't support integers as keys.

mirceaulinic:
@dbarrosop are you happy with this one? Can we merge and test with napalm-eos?

dbarrosop:
Yeah, haven't merged because I am still at RIPE's so I haven't had the time to fix that comment you mentioned but we can merge and test and fix the comments later on. Let me do it.

PR #124, opened 2016-11-10
Add method .get_vrfs()

bewing:
Proposal to add a method to retrieve configured VRFs and their associated interfaces from a device.

### DIFF
--- a/napalm_base/test/getters.py
+++ b/napalm_base/test/getters.py
@@ -425,3 +425,14 @@ class BaseTestGetters(object):
             assert get_config['running'] == "" if config != "running" else True
 
         return get_config
+
+    @wrap_test_cases
+    def test_get_network_instances(self):
+        """Test get_network_instances method."""
+        get_network_instances = self.device.get_network_instances()
+
+        assert isinstance(get_network_instances, dict)
+        for network_instance_name, network_instance in get_network_instances:
+            assert helpers.test_model(models.network_instance, network_instance)
+
+        return get_network_instances
--- a/napalm_base/test/getters.py
+++ b/napalm_base/test/getters.py
@@ -434,5 +434,8 @@ class BaseTestGetters(object):
         assert isinstance(get_network_instances, dict)
         for network_instance_name, network_instance in get_network_instances:
             assert helpers.test_model(models.network_instance, network_instance)
+            assert helpers.test_model(models.network_instance_state, network_instance['state'])
+            assert helpers.test_model(models.network_instance_interfaces,
+                                      network_instance['interfaces'])
 
         return get_network_instances
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1380,29 +1380,52 @@ class NetworkDriver(object):
         """
         raise NotImplementedError
 
-    def get_vrfs(self):
+    def get_network_instances(self, name=''):
         """
-        Return a dictionary of VRFs configured on the device, including default/global
-        The keys of the main dictionary represent the name of the VRF.  The values are as below:
-            * route_distinguisher (unicode or None)
-            * interfaces (list of unicodes)
+        Return a dictionary of network instances (VRFs) configured on the device, including default/global
+
+        Args:
+            name(string) - Name of the network instance to return, default is all.
+
+        Returns:
+            A dictionary of network instances in OC format:
+            * name (dict)
+              * name (unicode)
+              * type (unicode, fixed to L3VRF)
+              * state (dict)
+                * route_distinguisher (unicode or None)
+              * interfaces (dict)
+                * interface (dict)
+                  * interface name: (dict)
 
         Example:
         {
             u'MGMT': {
-                u'route_distinguisher': u'123:456',
-                u'interfaces': [
-                    u'Management1',
-                ]
+                u'name': u'MGMT',
+                u'type': u'L3VRF',
+                u'state': {
+                    u'route_distinguisher': u'123:456',
+                },
+                u'interfaces': {
+                    u'interface': {
+                        u'Management1': {}
+                    }
+                }
             }
             u'default': {
-                u'route_distinguisher': None,
-                u'interfaces: [
-                    u'Ethernet1',
-                    u'Ethernet2',
-                    u'Ethernet3',
-                    u'Ethernet4',
-                ]
+                u'name': u'default',
+                u'type': u'L3VRF',
+                u'state': {
+                    u'route_distinguisher': None,
+                },
+                u'interfaces: {
+                    u'interface': {
+                        u'Ethernet1': {}
+                        u'Ethernet2': {}
+                        u'Ethernet3': {}
+                        u'Ethernet4': {}
+                    }
+                }
             }
         }
         """
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1382,7 +1382,7 @@ class NetworkDriver(object):
 
     def get_network_instances(self, name=''):
         """
-        Return a dictionary of network instances (VRFs) configured on the device, including default/global
+        Return a dictionary of network instances (VRFs) configured, including default/global
 
         Args:
             name(string) - Name of the network instance to return, default is all.
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1379,3 +1379,31 @@ class NetworkDriver(object):
               empty string
         """
         raise NotImplementedError
+
+    def get_vrfs(self):
+        """
+        Return a dictionary of VRFs configured on the device, including default/global
+        The keys of the main dictionary represent the name of the VRF.  The values are as below:
+            * route_descriptor (unicode or None)
+            * interfaces (list of unicodes)
+
+        Example:
+        {
+            u'MGMT': {
+                u'route_descriptor': u'123:456',
+                u'interfaces': [
+                    u'Management1',
+                ]
+            }
+            u'default': {
+                u'route_descriptor': None,
+                u'interfaces: [
+                    u'Ethernet1',
+                    u'Ethernet2',
+                    u'Ethernet3',
+                    u'Ethernet4',
+                ]
+            }
+        }
+        """
+        raise NotImplementedError
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1384,19 +1384,19 @@ class NetworkDriver(object):
         """
         Return a dictionary of VRFs configured on the device, including default/global
         The keys of the main dictionary represent the name of the VRF.  The values are as below:
-            * route_descriptor (unicode or None)
+            * route_distinguisher (unicode or None)
             * interfaces (list of unicodes)
 
         Example:
         {
             u'MGMT': {
-                u'route_descriptor': u'123:456',
+                u'route_distinguisher': u'123:456',
                 u'interfaces': [
                     u'Management1',
                 ]
             }
             u'default': {
-                u'route_descriptor': None,
+                u'route_distinguisher': None,
                 u'interfaces: [
                     u'Ethernet1',
                     u'Ethernet2',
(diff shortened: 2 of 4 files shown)


mirceaulinic:
Hey @bewing,

It makes sense for me to have this method.

Few comments:

- when I firstly read the naming I expected it would return the list of configured vrfs. Personally I would find more clear if would be called ```get_vrf_config```
- by default return the config of all vrfs. We can have an optional arg (defaulting to None) that would select a specific vrf to return the config
- few other details such as route target, route distinguisher etc. would be quite useful

bewing:
* I didn't name it "get_vrf_config" as we're not actually parsing the running-config of the device, but instead parsing the output of the command that shows the VRFs on the device.  Don't know if this is semantics or not.
   * This seems like a good idea -- I didn't do it initially because my reference implementation CLI did not support "show vrf [name]", but it would be trivial to filter the dictionary before returning it if the optional arg is present.
   * Route targets are difficult, as you can have multiple per VRF, or may be set on a per-route basis by import-export maps.

dbarrosop:
I think this is a good idea. However, let me take a look at the openconfig model and propose something that aligns more with their model.

mirceaulinic:
@bewing merged https://github.com/napalm-automation/napalm-base/pull/125 - all green now.

> I didn't name it "get_vrf_config" as we're not actually parsing the running-config of the device, but instead parsing the output of the command that shows the VRFs on the device. Don't know if this is semantics or not.

Not sure if that's the right way - I will double check next week on other devices to confirm. I suppose you have reasons for eOS, but let me confirm to keep it uniform across platforms.

> Route targets are difficult, as you can have multiple per VRF, or may be set on a per-route basis by import-export maps.

Can't agree more. But if specified under that hierarchy, I would like to see it though. Otherwise, can be blank.

dbarrosop:
Ok, what about this:

Method name: `get_network_instances` (that's the name in OC land)

Object:

A dict of vrfs where the key is the name of the vrf and it looks like:

```
name: string
state: dict
  name:
  type: (fixed to L3VRF)
  route_distinguisher:
  interfaces: dict
    interface: dict of interfaces where the key is the interface name and the dict is empty
```

For example:

```
{
    u'MGMT': {
    	'name': 'MGMT',
    	'type': 'L3VRF',
    	'state': {
	        u'route_distinguisher': u'123:456',
	    }
        'interfaces': {
        	interface: {
            	u'Management1': {},
            }
        }
    }
    u'global': {
    	'name': 'global',
    	'type': 'L3VRF',
    	'state': {
	        u'route_distinguisher': None,
	    }
        'interfaces': {
        	interface: {
            	u'Ethernet1': {},
            	u'Ethernet2': {},
            	u'Ethernet3': {},
            }
        }
    }
}
```

Yeah, I know it's convoluted but that's how the portion we are extracting looks in the OC model so if we manage to move forward on the integration we are doing between napalm and OC we will be able to leverage this one : )

bewing:
That looks like a reasonable data structure to me.  So the only remaining question is the name of the method, which I'm agnostic on, as long as it doesn't lock us into a particular method for executing on a driver (config parsing versus output parsing).

dbarrosop:
> as long as it doesn't lock us into a particular method for executing on a driver

The name of the method doesn't have to match what it happens inside the driver. The only reason of choosing that method name is because that's how OC calls a VRF but that doesn't prevent you from doing `show vrf | json` on EOS, for example.

mirceaulinic:
With "few other details such as route target, route distinguisher etc. " I meant more interesting details (I thought the ```etc.``` would imply that) :)

The structure above does not expose enough details to convince me to use it. There's a ton of info to be returned, but here's a list of details I would like to see, at least:

- description
- import_policy
- export_policy
- routing_interface
- at least a list of protocols under the network instance
- route_target (I still vote for this - if not specified can be blank, does not hurt. And returning the policies, you know where to look to find it)

Also some other details such as router_id, ASN used would be quite good.

dbarrosop:
> import_policy
> export_policy

Those are not relevant to a VRF, maybe to a routing protocol attached to a VRF but not necessarily to the VRF itself.

> routing_interface

That's already there under `interfaces.interface`

> at least a list of protocols under the network instance

This one maybe I am fine with it.

> route_target

Couldn't find this construct on the OC specification...

mirceaulinic:
> import_policy
> export_policy
> Those are not relevant to a VRF, maybe to a routing protocol attached to a VRF but not necessarily to the VRF itself.

Yes, they are relevant (except for L2VPN) - did you ever see a multi VRF implementation without policies?
OC: https://github.com/openconfig/public/blob/master/release/models/network-instance/openconfig-network-instance.yang#L811

> routing_interface
> That's already there under `interfaces.interface`

Daaat! Nope, not the same.

> route_target
> Couldn't find this construct on the OC specification...

Indeed this is not. As @bewing pointed out this can be specified either under the VRF config, either added by the policy.

dbarrosop:
> Yes, they are relevant (except for L2VPN) - did you ever see a multi VRF implementation without policies?

ok, but that's the `route-target` then, isn't it?

> Daaat! Nope, not the same.

So... what do you mean then?

mirceaulinic:
> ok, but that's the route-target then, isn't it?

Correct, basically the route target is a shortcut to not mess with policies when the implementation is pretty straight. But when some of the routes have to be exported also in other VRFs, just specifying the target won't help. So depends on the complexity of the case. That's why I think it would be good to have both policies and target.

> So... what do you mean then?

The difference is tricky and the name kinda explains it, but not entirely. It is... the interface used for routing. This does not explain much, does it? Let me find some documentation, however this might be on Junos only.

dbarrosop:
I don't think OC has the `route-target` concept, we should bring it to them and make sure we are not missing anything.

> the interface used for routing

are you referring to IRB?

mirceaulinic:
> are you referring to IRB?

Yes

mirceaulinic:
> I don't think OC has the route-target concept, we should bring it to them and make sure we are not missing anything.

I did not find it either.

dbarrosop:
Ok, that's then the `fdb` in OC parlance, which is a different type of network-instance.

mirceaulinic:
> Ok, that's then the fdb in OC parlance, which is a different type of network-instance.

You mean this one: https://github.com/openconfig/public/blob/67b976166f9d4c873c034abf48feaf1925987226/release/models/network-instance/openconfig-network-instance-l2.yang#L74 right?

I thought this method should support both L2 and L3?

dbarrosop:
> I thought this method should support both L2 and L3?

Yes, everything. It's sort of overloaded in my opinion. The section you are highlighting is the `l2ni-instance` which is what includes the fdb which is the l2 part of it, which, by the way, looking better at it doesn't support the `routing-interface`. I got confused by something else. We should ask them I guess...

bewing:
> The structure above does not expose enough details to convince me to use it. There's a ton of info to be returned, but here's a list of details I would like to see, at least:

Knowledge of VRF names is primarily what I am going for here, as I'm looking forward to other VRF-aware commands that don't let you just say "all vrfs" / * like BGP does.  Eg. EOS "show ip pim vrf". 

> description

I could see a case for this, but it does destroy my reference implementation (show vrf on EOS doesn't show the description, you can only get it from config scraping).  Oh well, that shouldn't drive the decision anyway.

> import_policy
> export_policy

I'd definitely want clarification from the OC team on this.  It seems that this might be fueled by table-connections, but it seems Juniper-style intra-table protocol-specific there, instead of inter-instance leaking.

> route_target

I'm still against this, as a route target is associated with a route in MPBGP, and configuring them in VRFs is just shorthand import/export policy.

> routing_interface

In the current reference where only L3VRFs are defined, all interfaces in interfaces.interface are routing interfaces, are they not?

I also just noticed that OC does have a specific type for the default table, and will update to reflect.
https://github.com/openconfig/public/blob/67b976166f9d4c873c034abf48feaf1925987226/release/models/network-instance/openconfig-network-instance-types.yang#L73

mirceaulinic:
> I could see a case for this, but it does destroy my reference implementation (show vrf on EOS doesn't show the description, you can only get it from config scraping). Oh well, that shouldn't drive the decision anyway.

Maybe. But here we should not discuss how one would implement it for a specific operating system/driver, but a structure that would make sense for all.

> I'm still against this, as a route target is associated with a route in MPBGP, and configuring them in VRFs is just shorthand import/export policy.

Yes, agree with you, but say someone will set ```route-target 13335:1234``` on their device. Without this field the output looks like:

```python
{
   'route_distinguisher': 13335:1234,
   'import_policy': '',
   'export_policy': ''
   ... other fields ...
}
```

How can you tell looking at this how the routes will be imported/exported? Just considering this I would say that anything would be accepted, which does not reflect the reality.
But if the user decides to set the policies instead, the ```route_target``` field would be empty.
IMHO it is better to have more info than needed rather than lack of data.

> In the current reference where only L3VRFs are defined, all interfaces in interfaces.interface are routing interfaces, are they not?

No, for L2 they would not be.

bewing:
> How can you tell looking at this how the routes will be imported/exported? Just considering this I would say that anything would be accepted, which does not reflect the reality.

In your example, where you would have "route-target both 13335:1234", what should be done is 

```
table connection:
   * import_policy: from MP-BGP get all routes with extcomm RT 13335:1234
   * export_policy: to MP-BGP push all routes and add extcomm RT 1335:1234

```
I think we just need to clarify with OC that that is what was intended with regards to table connections, and how we want to express/display that information.

mirceaulinic:
> table connection:
   * import_policy: from MP-BGP get all routes with extcomm RT 13335:1234
   * export_policy: to MP-BGP push all routes and add extcomm RT 1335:1234

Both ```import_policy``` and ```export_policy``` fields should return the name of the policy/policies, not a story. BTW: on devices such as Junos you can set a list of policies - I guess they should be displayed here as space-separated values.

napalm_base/test/getters.py
```diff
@@ -425,3 +425,14 @@ def test_get_config_filtered(self):
...
         return get_config
+
+    @wrap_test_cases
+    def test_get_network_instances(self):
+        """Test get_network_instances method."""
+        get_network_instances = self.device.get_network_instances()
+
+        assert isinstance(get_network_instances, dict)
+        for network_instance_name, network_instance in get_network_instances:
+            assert helpers.test_model(models.network_instance, network_instance)
```
mirceaulinic: Can you please add one more nesting level and check also the keys under ```state``` and ```interfaces``` to make sure they have the right datatype?

This would imply adding two more models, e.g.: ```network_instance_state``` and ```network_instance_interfaces```.

But let's see also what @dbarrosop thinks about that.

bewing:
Hrm.  I think I was wrong on the table connections side -- it looks like inter-instance exchange is explicitly defined at https://github.com/openconfig/public/blob/master/release/models/network-instance/openconfig-network-instance.yang#L179

Which calls back to https://github.com/openconfig/public/blob/5960f079b1d15ecab867c702ad61867f5efcda10/release/models/policy/openconfig-routing-policy.yang

which I think is even more cans of worms.

mirceaulinic:
Oh god! Although this is one of my favourite topics in networking, indeed this is very complex and tons of variables. On the other hand I would really like this method to provide as many details as possible.

For the moment, I would propose to:

- decide whether & which of the fields I proposed would make sense to be added now. My take would be to have all ```import_policy```, ```export_policy``` and ```route_target``` returned if they have any content + ```description``` and few others from above
- later we can add others if needed, as would not be a detrimental change - would basically extend the output dictionary with one or several keys

Does this make sense?

dbarrosop:
I think we should merge this one as it is now and iterate on it if we think it's necessary. This method also serves as entry point to ask more information later on to other methods. We just need to add VRF support for them.

mirceaulinic:
Also the tests look good to me - for any further additions, we can revisit that later.

Thanks @bewing!

dbarrosop:
Has anyone updated the docs as well?

PR #128, opened 2016-11-16
New method: is_alive()

mirceaulinic:
There are cases when the connection is open correctly, the driver stays connected for minutes/hours/days, but due to external factors, the connection is dropped (packet loss, or the device simply kills the process etc.).
In that case, we need to know when the connection become unusable and reconnect.

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -89,6 +89,16 @@ class NetworkDriver(object):
         """
         raise NotImplementedError
 
+    def is_alive(self):
+        """
+        Returns a flag with the connection state.
+        Depends on the nature of API used by each driver.
+        The state does not reflect only on the connection status (when SSH), it must also take into
+        consideration other parameters, e.g.: NETCONF session might not be usable, althought the
+        underlying SSH session is still open etc.
+        """
+        raise NotImplementedError
+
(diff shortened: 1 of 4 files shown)


dbarrosop:
:facepalm:

You just expect either true or false, right? I'd still return

`{'is_alive': True}`

In case we decide later on add extra information about the liveness of the session.

mirceaulinic:
Makes sense, pushed the changes.

dbarrosop:
Looks good, we now need some tests : )

dbarrosop:
I stand corrected.

+1

mirceaulinic:
I initially have scheduled this one for 0.20.0. Should we alleviate it into 0.19.0 together with https://github.com/napalm-automation/napalm-base/pull/124?

dbarrosop:
I actually thins one might come in faster than the other one so I wouldn't be surprised if we end up swapping milestones ;)

mirceaulinic:
Okay, will insert this one into 0.19.0 then - IMO we could also ingest the other one for the moment and extend it later when needed.

dbarrosop:
> IMO we could also ingest the other one for the moment and extend it later when needed

I agree, that's the beauty of using dicts everywhere : )

PR #136, opened 2016-11-17
Implementation of get_firewall_policies

awlx:
It would be nice to have an output of all configured Firewall policies. In an easy and parseable way.

For this I would propose to use the method get_firewall_policies .

Output:
`>>> device.get_firewall_rules()
{'1': {'uuid': '9d747de8-dad4-51e5-7a7e-d078aa77140a', 'service': '"ALL"', 'schedule': '"always"', 'srcaddr': '"all"', 'dstintf': '"any"', 'srcintf': '"any"', 'action': 'accept', 'dstaddr': '"all"'}, '2': {'name': '"bla"', 'service': '"ALL"', 'schedule': '"always"', 'logtraffic': 'all', 'srcaddr': '"all"', 'dstintf': '"port2"', 'srcintf': '"port3"', 'dstaddr': '"all"', 'uuid': 'de597b64-aca8-51e6-5d00-b7874e6b72b8'}}`

I already implemented the feature for FortiOS:
https://github.com/napalm-automation/napalm-fortios/pull/22

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1452,27 +1452,19 @@ class NetworkDriver(object):
 
         Example:
         {
-            '1': {
-                'uuid': '9d747de8-dad4-51e5-7a7e-d078aa77140a',
-                'service': '"ALL"',
-                'schedule': '"always"',
-                'srcaddr': '"all"',
-                'dstintf': '"any"',
-                'srcintf': '"any"',
-                'action': 'accept',
-                'dstaddr': '"all"'
+            'src_zone-to-dst_zone': {
+                'position': 1,
+                'id': '230',
+                'enabled': True,
+                'schedule': 'Always',
+                'log': 'all',
+                'l3_src': 'any',
+                'l3_dst': 'any',
+                'service': 'HTTP',
+	            'src_zone': 'port2',
+            	'dst_zone': 'port3',
+	            'action': 'Permit'
             }
-            '2': {
-                'name': '"bla"',
-                'service': '"ALL"',
-                'schedule': '"always"',
-                'logtraffic': 'all',
-                'srcaddr': '"all"',
-                'dstintf': '"port2"',
-                'srcintf': '"port3"',
-                'dstaddr': '"all"',
-                'uuid': 'de597b64-aca8-51e6-5d00-b7874e6b72b8'
-           }
         }
         """
         raise NotImplementedError
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1443,8 +1443,8 @@ class NetworkDriver(object):
 
     def get_firewall_policies(self):
         """
-        Returns a dictionary of lists of dictionaries where the first key is an unique policy 
-	name and the inner dictionary contains the following keys:
+        Returns a dictionary of lists of dictionaries where the first key is an unique policy
+        name and the inner dictionary contains the following keys:
 
         * position (int)
         * id (text_type)
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1461,9 +1461,9 @@ class NetworkDriver(object):
                 'l3_src': 'any',
                 'l3_dst': 'any',
                 'service': 'HTTP',
-	            'src_zone': 'port2',
-            	'dst_zone': 'port3',
-	            'action': 'Permit'
+                'src_zone': 'port2',
+                'dst_zone': 'port3',
+                'action': 'Permit'
             }
         }
         """
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1440,3 +1440,39 @@ class NetworkDriver(object):
         }
         """
         raise NotImplementedError
+
+    def get_firewall_policies(self):
+        """
+        Return a dictionary of configured Firewall policies
+
+        Returns:
+            A dictionary of policies 
+            * name
+                * n amount of arguments
+
+        Example:
+        {
+            '1': {
+                'uuid': '9d747de8-dad4-51e5-7a7e-d078aa77140a',
+                'service': '"ALL"', 
+                'schedule': '"always"', 
+                'srcaddr': '"all"', 
+                'dstintf': '"any"', 
+                'srcintf': '"any"', 
+                'action': 'accept', 
+                'dstaddr': '"all"'
+            } 
+            '2': {
+                'name': '"bla"',
+                'service': '"ALL"',
+                'schedule': '"always"', 
+                'logtraffic': 'all',
+                'srcaddr': '"all"', 
+                'dstintf': '"port2"', 
+                'srcintf': '"port3"', 
+                'dstaddr': '"all"', 
+                'uuid': 'de597b64-aca8-51e6-5d00-b7874e6b72b8'
+           }
+        }
+        """
+        raise NotImplementedError 
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1456,7 +1456,7 @@ class NetworkDriver(object):
         * service (text_type)
         * src_zone (text_type)
         * dst_zone (text_type)
-        * action (text_type)        
+        * action (text_type)
 
         Example::
 
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1443,7 +1443,7 @@ class NetworkDriver(object):
 
     def get_firewall_policies(self):
         """
-        Returns a dictionary of dictionaries where the first key is an unique policy name and the
+        Returns a dictionary of lists of dictionaries where the first key is an unique policy name and the
         inner dictionary contains the following keys:
 
         * position (int)
@@ -1461,7 +1461,7 @@ class NetworkDriver(object):
         Example::
 
         {
-            'src_zone-to-dst_zone': {
+            'policy_name': [{
                 'position': 1,
                 'id': '230',
                 'enabled': True,
@@ -1473,7 +1473,7 @@ class NetworkDriver(object):
                 'src_zone': 'port2',
                 'dst_zone': 'port3',
                 'action': 'Permit'
-            }
+            }]
         }
         """
         raise NotImplementedError
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1446,7 +1446,7 @@ class NetworkDriver(object):
         Return a dictionary of configured Firewall policies
 
         Returns:
-            A dictionary of policies 
+            A dictionary of policies
             * name
                 * n amount of arguments
 
@@ -1475,4 +1475,4 @@ class NetworkDriver(object):
            }
         }
         """
-        raise NotImplementedError 
+        raise NotImplementedError
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1443,14 +1443,23 @@ class NetworkDriver(object):
 
     def get_firewall_policies(self):
         """
-        Return a dictionary of configured Firewall policies
+        Returns a dictionary of dictionaries where the first key is an unique policy name and the
+        inner dictionary contains the following keys:
 
-        Returns:
-            A dictionary of policies
-            * name
-                * n amount of arguments
+        * position (int)
+        * id (text_type)
+        * enabled (bool)
+        * schedule (text_type)
+        * log (text_type)
+        * l3_src (text_type)
+        * l3_dst (text_type)
+        * service (text_type)
+        * src_zone (text_type)
+        * dst_zone (text_type)
+        * action (text_type)        
 
-        Example:
+        Example::
+
         {
             'src_zone-to-dst_zone': {
                 'position': 1,
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1447,6 +1447,8 @@ class NetworkDriver(object):
         name and the inner dictionary contains the following keys:
 
         * position (int)
+        * packet_hits (int)
+        * byte_hits (int)
         * id (text_type)
         * enabled (bool)
         * schedule (text_type)
@@ -1463,6 +1465,8 @@ class NetworkDriver(object):
         {
             'policy_name': [{
                 'position': 1,
+                'packet_hits': 200,
+                'byte_hits': 83883,
                 'id': '230',
                 'enabled': True,
                 'schedule': 'Always',
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1454,23 +1454,23 @@ class NetworkDriver(object):
         {
             '1': {
                 'uuid': '9d747de8-dad4-51e5-7a7e-d078aa77140a',
-                'service': '"ALL"', 
-                'schedule': '"always"', 
-                'srcaddr': '"all"', 
-                'dstintf': '"any"', 
-                'srcintf': '"any"', 
-                'action': 'accept', 
+                'service': '"ALL"',
+                'schedule': '"always"',
+                'srcaddr': '"all"',
+                'dstintf': '"any"',
+                'srcintf': '"any"',
+                'action': 'accept',
                 'dstaddr': '"all"'
-            } 
+            }
             '2': {
                 'name': '"bla"',
                 'service': '"ALL"',
-                'schedule': '"always"', 
+                'schedule': '"always"',
                 'logtraffic': 'all',
-                'srcaddr': '"all"', 
-                'dstintf': '"port2"', 
-                'srcintf': '"port3"', 
-                'dstaddr': '"all"', 
+                'srcaddr': '"all"',
+                'dstintf': '"port2"',
+                'srcintf': '"port3"',
+                'dstaddr': '"all"',
                 'uuid': 'de597b64-aca8-51e6-5d00-b7874e6b72b8'
            }
         }
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1443,8 +1443,8 @@ class NetworkDriver(object):
 
     def get_firewall_policies(self):
         """
-        Returns a dictionary of lists of dictionaries where the first key is an unique policy name and the
-        inner dictionary contains the following keys:
+        Returns a dictionary of lists of dictionaries where the first key is an unique policy 
+	name and the inner dictionary contains the following keys:
 
         * position (int)
         * id (text_type)
--- a/napalm_base/test/base.py
+++ b/napalm_base/test/base.py
@@ -166,6 +166,14 @@ class TestGettersNetworkDriver(object):
 
         return correct_class and same_keys
 
+    def test_get_firewall_policies(self):
+        try:
+            policies = self.device.get_firewall_policies()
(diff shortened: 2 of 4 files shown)


napalm_base/base.py
```diff
@@ -1440,3 +1440,39 @@ def get_network_instances(self, name=''):
...
+            '2': {
+                'name': '"bla"',
+                'service': '"ALL"',
+                'schedule': '"always"',
+                'logtraffic': 'all',
+                'srcaddr': '"all"',
+                'dstintf': '"port2"',
+                'srcintf': '"port3"',
+                'dstaddr': '"all"',
+                'uuid': 'de597b64-aca8-51e6-5d00-b7874e6b72b8'
```
dbarrosop: I think we should ignore the `uuid`, for two reasons:

1. It's a fortios construct not present on any other vendor.
2. `pyfg` ignores that field anyway  :)

napalm_base/base.py
```diff
@@ -1440,3 +1440,39 @@ def get_network_instances(self, name=''):
...
+        Return a dictionary of configured Firewall policies
+
+        Returns:
+            A dictionary of policies
+            * name
+                * n amount of arguments
+
+        Example:
+        {
+            '1': {
```
dbarrosop: I think we should make this a bit more vendor agnostic. What about this?

```
{
   'my_awesome_policy_name': [
                 {
                'id': 1,   # position of this entry in the policy
                'enabled': True,  # is this entry active or not?
                'schedule': [],   # when it's enabled, a list of {'start_time': datetime, 'end_time': datetime}. If empty, it means it's not enabled.
                'log': true,  # is traffic logged?
                'l3_src': "any",  # string that should point to some address book type object
                'l3_dst':  "any",  # string that should point to some address book type object
                'service':  "http",  # string that should point to some address book type object
                'src_zone': "port2",  # string that should point to an interface or security zone or whatever construct the device uses. `None` if the vendor doesn't use this construct, `any` if it's used but it doesn't matter.
                'dst_zone': "port3", # same as src_zone
                'action': "permit" # permit, reject (send ICMP destination-unreachable back), drop (silently discard traffic)
                 },
                 {
                'id': 100,   # position of this entry in the policy
                'enabled': True,
                'schedule': [],
                'log': true,  # is traffic logged?
                'l3_src': "any",
                'l3_dst':  "any",
                'service':  "any",
                'src_zone': "port2",
                'dst_zone': "port3",
                'action': "drop" # permit, reject, drop
                 },
 ]
}
```

DiogoAndre:
This is looking good!  A couple of comments:

> `'id': 1,   # position of this entry in the policy`

I think we could go with ```position``` instead of ```id``` for this purpose.  The rule may have a separate ID other than the position in the list, and this ID sometimes can be useful for getting more information about the entry and/or correlate with information form another command or api endpoint.

But we should also keep the ID and populate it with the provided ID for the entry, if available.

> `'service':  "http",  # string that should point to some address book type object`

Would this be something like:

```python
"http": {
  "protocol": "tcp",
  "port": "80"
}
``` 
?

On a diffrente topic, I also wonder if it would be possible to include some other information like the number of hits on the rule, or a timestamp of when it was last used. I think some sort of `hitcount` is available on most platforms, but not sure about a timestamp.

dbarrosop:
> `position` and `id`

Ok, the original proposal had the ID as well so I guess the people has spoken ;) I have edited the previous message to reflect those two fields.

> Would this be something like:

No, just a string. AFAIK most firewalls attach to firewall rules some sort of service name that gets translated internally on some address book into a set of `protocols/ports`. So here we would just return the string. If needed, we could have another method to read the address books and match those. If a system doesn't work like this (for example, a simple ACL) I guess we could just encode the info in the string, ie, `"tcp/80"` instead of just `"http"`.

> timestamp of when it was last used. I think some sort of hitcount

I think `hits` might make sense. @awlx what do you think?

awlx:
hits would make sense but I am not sure if for example Fortinet supports that. It only has `"Bytes matched"`. The `my_fancy_policy_name` is also something Fortinet does not have. 

The nameing of all fields are just unique to Fortinet atm, which is clearly not very good for every platform. So I agree we need to rework that.

For Fortinet you can also define your own services for example which would be looked up via Fortinet CLI to generate a mapping.

dbarrosop:
> hits would make sense but I am not sure if for example Fortinet supports that. It only has "Bytes matched"

I think we should have two fields, `packet_hits` and `byte_hits` and just populate whatever it's available and set to -1 the other. Thoughts?

> The my_fancy_policy_name is also something Fortinet does not have

Yeah, I suggest you to just build some name dynamicaly, something like `"from_zoneA_to_any"` or whatever. It will not have much meaning other than consistency across drivers.

> For Fortinet you can also define your own services for example which would be looked up via Fortinet CLI to generate a mapping.

Same for SRXs and I suspect all the vendors in general. That would be a different method though.

DiogoAndre:
> I think we should have two fields, packet_hits and byte_hits and just populate whatever it's available and set to -1 the other. Thoughts?

👍

dbarrosop:
Great, I think we got here to an understanding : )

Unless @awlx has a last hour comment I will let her adapt her PR here : )

Thanks!

awlx:
Ok, I have implemented the format we discussed in the Fortinet Provider. But before I push it, what do you think? :)

`{
'port3-to-port2':
	 {
     'src_zone': 'port3',
	 'log': 'all',
	 'service': 'ALL',
	 'schedule': 'always',
	 'l3_src': 'all',
	 'enabled': True,
	 'l3_dst': 'all',
	 'dst_zone': 'port2',
	 'action': 'reject',
	 'position': 1, 
         'id': '2'
         },
 'any-to-any':
 	{
	'src_zone': 'any',
	 'log': False,
	 'service': 'ALL',
	 'schedule': 'always',
	 'l3_src': 'all',
	 'enabled': False, 
	 'l3_dst': 'all',
	 'dst_zone': 'any',
	 'action': 'permit',
	 'position': 0,
	 'id': '1'
      }
}`

napalm_base/test/base.py
```diff
@@ -166,6 +166,14 @@ def _test_model(model, data):
 
         return correct_class and same_keys
 
+    def test_get_firewall_policies(self):
+        try:
+            policies = self.device.get_firewall_policies()
+        except NotImplementedError:
+            raise SkipTest()
+        result = len(policies["any-to-any"]) > 0
```
mirceaulinic: This should not be hardcoded. Please define the model, specifying the keys and their datatype, e.g.: https://github.com/napalm-automation/napalm-base/blob/develop/napalm_base/test/models.py#L20-L27
Afterwards, you'll need an iteration and verify if they respect, e.g.: https://github.com/napalm-automation/napalm-base/blob/develop/napalm_base/test/base.py#L185-L195

awlx: No longer, hardcoded. Sorry, used this for an internal test and forgot to remove :).

mirceaulinic: Mmm... I'm afraid that I asked for more than that - did you have look at the models and the examples referenced above?

You'll need to:

1. define a new model
2. write tests using that model

Let us know if you'd need more info on this.

awlx: How do I represent the `fancy_policy_name` being a dict() and the other items inside of that?

awlx: Thx, will do that.

napalm_base/test/getters.py
```diff
@@ -463,3 +463,15 @@ def test_get_network_instances(self, test_case):
                                       network_instance['interfaces'])
 
         return get_network_instances
+    
+    @wrap_test_cases
+    def test_get_firewall_policies(self, test_case):
+        """Test get_firewall_policies method."""
+        get_firewall_policies = self.device.get_firewall_policies()
+        assert len(get_firewall_policies) > 0
+        for policy in get_firewall_policies:
```
mirceaulinic: Please replace with:

```python
for policy_name, policy_details in get_firewall_policies.items():
    assert helpers.test_model(models. firewall_policies, policy_details)
```

To use the model you just defined.

mirceaulinic:
@awlx pylama is complaining: ```The command "pylama ." exited with 1.```

napalm_base/base.py
```diff
@@ -1440,3 +1440,31 @@ def get_network_instances(self, name=''):
...
+    def get_firewall_policies(self):
+        """
+        Return a dictionary of configured Firewall policies
+
+        Returns:
+            A dictionary of policies
+            * name
+                * n amount of arguments
+
+        Example:
```
mirceaulinic: ```Example::``` instead of ```Example:``` and newline after

napalm_base/base.py
```diff
@@ -1440,3 +1440,31 @@ def get_network_instances(self, name=''):
...
         """
         raise NotImplementedError
+
+    def get_firewall_policies(self):
+        """
+        Return a dictionary of configured Firewall policies
+
+        Returns:
+            A dictionary of policies
+            * name
```
mirceaulinic: Please provide more info here - say exactly what fields we should expect. For example, you can follow the pattern from ```get_interfaces_counters ``` https://github.com/napalm-automation/napalm-base/blob/develop/napalm_base/base.py#L357

awlx:
I think we should also add the `position` to `my_fancy_policyname`because otherwise it gets overwritten when more than one rule exists for src_int/dst_int. Is `position`a good idea? Or something else?

mirceaulinic:
> I think we should also add the position to my_fancy_policyname

What is "my_fancy_policyname"? If that's the key, no - this won't work for other drivers. I guess that position detail is specific to Fortios only. The name of the policy should be unique.

awlx:
At the moment I build it of the combination of `source-interface-to-destination-interface`. Which makes sense for all platforms. But we need something to make it unique. So what could be the perfect match? Position shouldn't be specific for Fortinet, because you always have a specific position inside the ruleset? Like this is the "5th" rule in the ACL.

DiogoAndre:
Hey... I think a small detail is causing this confusion. The structure is not to be directly a _dictionary of dictionaries_  but a _dictionary of **lists** of dictionaries_

So instead of having:

```python
{
            'src_zone-to-dst_zone': {
                'position': 1,
                'id': '230',
                'enabled': True,
                'schedule': 'Always',
                'log': 'all',
                'l3_src': 'any',
                'l3_dst': 'any',
                'service': 'HTTP',
                'src_zone': 'port2',
                'dst_zone': 'port3',
                'action': 'Permit'
            }
}
```

we should have 

```python
{
            'src_zone-to-dst_zone': [{
                'position': 1,
                'id': '230',
                'enabled': True,
                'schedule': 'Always',
                'log': 'all',
                'l3_src': 'any',
                'l3_dst': 'any',
                'service': 'HTTP',
                'src_zone': 'port2',
                'dst_zone': 'port3',
                'action': 'Permit'
            }]
}
```

The important detail here is that the `src_zone-to-dst_zone` key can now contain multiple dictionaries describing the individual rules.

There is a good example of a similar structure in the [get_lldp_neighbors_detail](https://github.com/napalm-automation/napalm-base/blob/develop/napalm_base/base.py#L424) method implementation.

dbarrosop:
Yes, @DiogoAndre is completely right.

awlx:
I am not sure if this is the best way to represent this. Maybe we should use the ID as another dict inside the `srczone-to-destzone`.
`{
            'src_zone-to-dst_zone': {
             	    '230': {
               		 'position': 1,
               		 'enabled': True,
                	 'schedule': 'Always',
                	 'log': 'all',
                	 'l3_src': 'any',
                	 'l3_dst': 'any',
               		 'service': 'HTTP',
              	         'src_zone': 'port2',
                         'dst_zone': 'port3',
                         'action': 'Permit'
           	    }
	    }
}`

With this we could clearly identify the individual rules. And would help us to further process the rules. And make changes to one specific line in the ACL.

dbarrosop:
> Maybe we should use the ID 

What do the rest thing about that? I have no particular opinion. However, AFAIK the only platform that has a sort of id for individual entries is fortios (the uuid), the rest they just have a position. We could use that as the id but then I want the test to make sure we are using strings as keys. Something like: `assert all([isinstance(k, int) for k in result.keys())`

awlx:
I don't use the UUID for building the ID. The ID is atm the rule number (not the position), which should be available on all platforms. Fortinet has UUID, ID, Position. Where ID is the one to actually identify a rule inside the ruleset. Maybe we for other platforms we just use the `position` as ID.

dbarrosop:
Let's see if @DiogoAndre @mirceaulinic have a particular opinion about using a dict or a list. I don't as long as we ensure keys are strings :)

mirceaulinic:
As @DiogoAndre said, a list would make more sense to me too.

awlx:
But how can I access a specific rule with all of it's settings related to it then? Because with a list I don't know which rule is the one with HTTP Service allowed for example.

dbarrosop:
How do you know by just looking at the ID which is a completely random and meaningless string? In any case, you could do:

```
for p in d.get_firewall_policies():
   if p['id'] == id:
       my_policy = p
```

awlx:
I thought of ID as the unique identifier for one exact rule. As I said might be the same as position on some platforms. But the actual ID on Fortinet.

dbarrosop:
Do you have an example of what the ID is on a fortinet device? Because I think we are lost in translation here : )

In any case, we have to aim something that makes sense for the majority of the platforms, not for one in particular. But let's see what you mean and then evaluate.

awlx:
See above :).

mirceaulinic:
@awlx in napalm-base we should only focus on a structure that makes sense on all platforms.
In your particular case - Fortinet - when you jump on the CLI and look at the config how do you uniquely identify the policies?

dbarrosop:
'230'? How do know what that rule does? I mean, if you know it already, why do you even need this method? (sorry that was half a real question half trolling :P)

awlx:
> @awlx in napalm-base we should only focus on a structure that makes sense on all platforms.
In your particular case - Fortinet - when you jump on the CLI and look at the config how do you uniquely identify the policies?
Via the "edit 230" for example. That's a unique rule at Fortinet CLI and describes exact one rule.

Yes, I understand that we want a structure that works on all platforms. I am just asking if the structure really makes sense. Or if we should change the schema a bit to make accessing single rules easier. For example to later add the features to manipulate them.

mirceaulinic:
> I am just asking if the structure really makes sense.

For me it makes sense - the policy name is the first key in the output and is unique. If on fortinet you don't have it, we can discuss there under the `napalm-fortinet` driver how to build an unique key to identify the policy.
Each policy can have multiple terms, at different positions, thus the values under the policy should be lists, following David's and Diogo's suggestions.

DiogoAndre:
There are probably some differences on how policies are created and applied between different vendors. I think having a list of rules makes more sense because the ID might not always be present, or not have the same meaning across the different implementations (i.e. the ID available in Cisco is more akin to the UUID in Fortinet)

awlx:
Ok, I had a fail in thinking sorry. Was too focused on Fortinet (that's what you get from working with that every day ;)). 

I will change the code to a list. And will change the Fortinet Driver to use the unique ID as the policyname, as this is what it is.

dbarrosop:
Great, thx! This is why this discussion are useful, to try to come up with something more generic. We all know very well our platforms so we are full of biases : )

ogenstad:
How complete should this be?

Other things to consider might be:

* source ports
* ranges
* Cisco style standard access-lists
* layer 2 access-lists
* Web filters
* Inline address translation

There are also fun stuff like Zone Based Firewall on some Cisco devices which work completely different, I'm sure other vendors have something similar.

mirceaulinic:
Indeed there are many things indeed to be considered. In addition to what @ogenstad suggested, I would personally like to see also two more fields: ```source_prefix_list``` and ```destination_prefix_list``` and potentially others later.

dbarrosop:
> source ports

That's true, maybe it's worth having src_ports and dst_ports, both pointing to a string instead of just service.

> ranges

That's covered IMHO. Major fields are strings that are supposed to be strings.

> Cisco style standard access-lists

That's an artificial construct (leave unused fields empty). And to be honest, it's not a firewall policy.

> layer 2 access-lists

Out-of-scope?

> Web filters

Out-of-scope?

> Inline address translation

Do you mean NAT?

I hate firewalls... they just make things difficult...

dbarrosop:
> source_prefix_list and destination_prefix_list

How is that different from what we have already?

mirceaulinic:
```souce_prefix_list``` and ```destination_prefix_list``` would be the names of the prefix lists referenced. Is this covered already by other fields?

**Comment by dbarrosop** [comment]
Why wouldn't you put those in the fields 'l3_src' and 'l3_dst'?

**Comment by mirceaulinic** [comment]
Yes, that's right.

File: napalm_base/test/base.py [hunk 7]
```diff
@@ -166,6 +166,19 @@ def _test_model(model, data):
...
 
+    def test_get_firewall_policies(self):
+        try:
+            policies = self.device.get_firewall_policies()
+        except NotImplementedError:
+            raise SkipTest()
+
+        result = len(policies) > 0
+
+        for policy_name, policy_details in policies.items():
```
mirceaulinic: Given that each value is a list now, you should adapt the tests to:

```python
for policy_name, policy_details in policies.items():
    for policy_term in policy_details:
        result = result and self._test_model(models.firewall_policies, policy_term)
```

napalm_base/test/getters.py
```diff
@@ -463,3 +463,13 @@ def test_get_network_instances(self, test_case):
                                       network_instance['interfaces'])
 
         return get_network_instances
+
+    @wrap_test_cases
+    def test_get_firewall_policies(self, test_case):
+        """Test get_firewall_policies method."""
+        get_firewall_policies = self.device.get_firewall_policies()
+        assert len(get_firewall_policies) > 0
+        for policy_name, policy_details in get_firewall_policies.items():
```
mirceaulinic: Same reasoning here:

```python
for policy_name, policy_details in get_firewall_policies.items():
    for policy_term in policy_details:
        assert helpers.test_model(models.firewall_policies, policy_term)
```

awlx: Fixed

napalm_base/test/base.py
awlx: Fixed

napalm_base/base.py
```diff
@@ -1440,3 +1440,40 @@ def get_network_instances(self, name=''):
...
+
+        Example::
+
+        {
+            'policy_name': [{
+                'position': 1,
+                'id': '230',
+                'enabled': True,
+                'schedule': 'Always',
+                'log': 'all',
```
awlx: The "log" field? I don't think so as most platforms support, "all" "utm only" "only denies" "disabled" and other messages. So I would prefer a text_type.

PR #139, opened 2016-11-18
Fix #138: Errors in get_network_instances

### LINKED ISSUE #138: Errors in get_network_instances
There is an issue in the model:  route_descriptor is incorrectly referenced instead of route_distinguisher.

In the test, we incorrectly attempt to iterate over the keys instead of key/value.

bewing:


mirceaulinic:
Can you fix also the docstring in https://github.com/napalm-automation/napalm-base/blob/develop/napalm_base/base.py#L1406 - should be just ```unicode```

PR #157, opened 2016-12-05
New test for routes with no next-hop IP

bewing:
In order to test new code in napalm-automation/napalm-eos#98 a new test case is needed to use the non-BGP codepath.


dbarrosop:
We need to implement that test on the rest of the drivers to implement this method or CI will break. I will wait until we have linked PRs implementing this test for those here.

For reference, those drivers are: EOS,	JunOS and IOS-XR

dbarrosop:
Mmm, can you rebase? You are somehow introducing commits that are not yours and that are actually merged already.

bewing:
I have no idea how those commits got in there, but they're also in my develop branch, which is now showing as ahead of upstream/develop.  I may have to nuke this PR and start over. :/

PR #160, opened 2016-12-07
Fixing issues with Python3 and method signature check

ktbyers:
Python3 inspect.ismethod() requires a bound object.

Consequently, the following line was always reporting as True (and the method signature checks were being skipped).

```
if attr.startswith('_') or not inspect.ismethod(func):
    continue
```

### DIFF
--- a/napalm_base/test/getters.py
+++ b/napalm_base/test/getters.py
@@ -116,27 +116,31 @@ class BaseTestGetters(object):
         """Test that all methods have the same signature."""
         errors = {}
         cls = self.driver
-        attrs = [m for m in dir(cls)]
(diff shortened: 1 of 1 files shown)


ktbyers:
Please test the above against a working NAPALM driver (i.e. one that has the correct signatures).

I was testing this against napalm-ios and its signatures are failing (and this test properly detects this).

ktbyers:
This worked for napalm-ios for both Python2 and Python3 (after I fixed the incorrect signatures).

Still might want to regression test it against a different napalm driver.

dbarrosop:
Great! I tested with `napalm-eos` and it worked great.

Thanks for the fixes : )

PR #100, opened 2016-09-18
Adding validation capabilities to the napalm-base

GGabriele:
Refers to #83

Here I'm adding validation capabilities to napalm-base. This means that we'll be able to validate getters method results and running/candidate configuration as well.

Sample `validate.yml`

```

---

get_interfaces:
  Ethernet2/5:
    is_enabled: true
    is_up: false

  loopback103:
    is_enabled: false
    is_up: false

  Vlan100:
    is_enabled: false
    is_up: false

get_facts:
  os_version: 7.0(3)I2(2d)
  interface_list:
    - Vlan5
    - Vlan100
  hostname: n9k2

get_lldp_neighbors:
  mgmt0:
    - hostname: PERIMETER
      port: Fa1/0/3
    - hostname: nxos-spine2
      port: nxos-spine2

get_bgp_neighbors:
  default:
    router_id: 192.0.2.2
    peers:
      192.0.2.3:
        is_enabled: false
      192.0.2.2:
        is_enabled: false
        address_family:
          ipv4:
            sent_prefixes: 5
          ipv6:
            sent_prefixes: 2

get_arp_table:
  - interface: Ethernet2/1
    ip: 192.0.2.2

get_interfaces_ip:
  Ethernet2/1:
    ipv4:
      192.0.2.1:
        prefix_length: 30

running:
  - username gabriele password gabriele
  - interface Ethernet2/1
```

Sample implementation into NXOS driver:

```
    def validate(self, getters=None, config=None, validation_file='validate.yml'):
        errors = napalm_base.validate.validate(self, getters=getters, config=config, validation_file=validation_file)
        return errors
```

Sample workflow:

```
>>>import json
>>>from napalm import get_network_driver
>>>driver = get_network_driver('nxos')
>>>device = driver('nxos', 'ntc', 'ntc123')
>>>device.open()
>>>
>>>getters = ['get_facts', 'get_lldp_neighbors', 'get_bgp_neighbors', 'get_interfaces', 'get_interfaces_ip']
>>>validation = device.validate(getters=getters)
>>>
>>>print json.dumps(validation, indent=4)
```

Sample results:

```
{
    "get_facts": {
        "os_version": [
            "Expected key but not found."
        ], 
        "interface_list": [
            "Expected but not found: Vlan5", 
            "Expected but not found: Vlan100"
        ], 
        "hostname": [
            "Expected 'n9k2', found 'nxos-spine1' instead"
        ]
    }, 
    "get_interfaces": {
        "Ethernet2/5": [
[...]

PR #161, opened 2016-12-10
Adding validation capabilities to the napalm-base (#100)

dbarrosop:
Do NOT merge, this is mostly to discuss features and collaborate.

TODO:

1. A CLI tool.
2. Ansible module for assertions (TBD after this is actually released).

Quick how to:

The idea of this feature is to use napalm to build compliance reports. Given a validation file like this:

https://github.com/napalm-automation/napalm-base/blob/validate/test/unit/validate/mocked_data/simple/validate.yml

Napalm is going to retrieve data and configuration from the live device and verify it against the file. Then, in your code you could do:

```
with driver(hostname, username, ...) as device:
    errors = device.compliance_errors("/path/to/compliance.yml")
    if not errors:
          print("We are compliant!!!)
    else:
          print("We are not:\n{}".format(errors))
```

And get things like:

```
# if the device is compliant
{}

# If the device has some live data that doesn't comply
{'get_arp_table': {'not_matched': [
                    "Expected but not found: {'interface': 'Ethernet3/1', 'ip': '192.0.2.3'}"]},
 'get_bgp_neighbors': {'default': ['Expected key but not found.']}}

# Or even configuration
{'Expected but not found in running config': ['logging server 10.0.0.1']}
```


dbarrosop:
@GGabriele I have done some "minor" changes:

1. No need of passing a list of getters to run anymore.
2. Instead of returning `true` if everything is fine or a dict with the errors, we always return a dict with the errors. If it's empty, it means everything is fine.
3. I renamed the method to `compliance_errors` as it's more readable given the return value of the function.

Thoughts?

GGabriele:
This looks good to me, great work.

Regarding the CLI tool, where should we push it? From a CLI perspective, how do you want this to look?

dbarrosop:
@GGabriele there are some here already: https://github.com/napalm-automation/napalm-base/tree/develop/napalm_base/clitools

GGabriele:
@dbarrosop I can help with that when you'll merge this if you want

dbarrosop:
Guys, I am going to merge this. A couple of notes:

1. The configuration validation part is out for the moment. I want to see if we can integrate it somehow with nelkit
2. Doc is published here: https://napalm.readthedocs.io/en/latest/validate/index.html

dbarrosop:
@GGabriele feel free to take a look to the CLI, let me know if you need help : )

PR #162, opened 2016-12-10
Test case for connected routes

bewing:
I don't know how napalm-automation/napalm-base#157 got so horribly screwed, so I just deleted everything and started over.

### DIFF
--- a/napalm_base/test/getters.py
+++ b/napalm_base/test/getters.py
@@ -358,6 +358,22 @@ class BaseTestGetters(object):
         return get_route_to
 
     @wrap_test_cases
+    def test_get_route_to_connected(self, test_case):
+        """Test get_route_to."""
+        destination = '1.0.4.0/24'
+        protocol = 'connected'
+        get_route_to_connected = self.device.get_route_to(destination=destination,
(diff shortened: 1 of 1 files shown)


mirceaulinic:
Putting this on hold till we'll have the implementation available on all drivers.

dbarrosop:
Updated the description of the PR adding a list of devices we have to implement this for. We can create issues if we prefer and link them here if we prefer.

manuel-domke:
Hi,
I just added a IPv6 implementation to get_route_to in the iosxr driver. 

Now **need** to extend this testcase to IPv6 destinations for my changes to pass the CI.
Any comments on that?

dbarrosop:
This other PR, which is similar to this one, made me think:

https://github.com/napalm-automation/napalm-base/pull/233

Quoting myself from that other PR: "I was wondering if we should just remove the "protocol" from test_get_route_to so we can control with the test cases (default and the rest) the data we get and parse instead."

dbarrosop:
Please, see: https://github.com/napalm-automation/napalm-base/issues/238

PR #163, opened 2016-12-10
add support napalm conf file

jedelman8:
This PR is to simplify the creation of NAPALM device objects using a NAPALM configuration file.  Doesn't change the way anything works - just offers this as an option.

The current approach:

```
from napalm import get_network_driver
driver = get_network_driver('ios')
device = driver('csr1', 'username', 'password')

```

With this PR, you will also be able to do:

```
from napalm import network_device
device = network_device('csr1')
```

Requires this: https://github.com/napalm-automation/napalm/pull/328

`network_device` is also opening the connection, i.e. calling the `open` method.  The conf file is also nice as it hides (abstracts) details such as username, password, and optional args per device.

Currently, any parameter that is NOT `username`, `password`, `hostname`, or `timeout` is considered an *optional_arg* and will be passed to the NetworkDriver as such.

This is a sample Napalm configuration file:

```
[nxos:n9k1]
username: ntc
password: ntc123

[nxos:n9k2]
username: ntc
password: ntc123

[ios:csr1]
hostname: 1.1.1.1
username: ntc
password: ntc123

[eos:eos-spine1]
hostname: 2.2.2.2
username: ntc
password: ntc123


```

The structure is:  `[<napalm_os_type>:<device_name>]` where `napalm_os_type` should match a given driver while `device_name` can either be a IP/FQDN or just a device reference.  If the `hostname` key is not given for a device, the `device_name` is used as the `hostname` in the NetworkDriver.

There is currently no support for groups and defaults. Each device must be defined.  Not difficult if you template out this file.

The file can be specified one of three ways and in this order:
- use the `filename` parameter when instantiating a new device object with this new method called `network_device`
- use the NAPALM_CONF environment variable
- use `~/napalm.conf` (in your home directory)

 
TODO:
  - Still need to add custom exceptions
  - Wanted feedback first if it makes sense
  - Any guidance on tests for this would be helpful too

### DIFF
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -25,10 +25,7 @@ import inspect
 import importlib
 import pkg_resources
 
-try:
-    from configparser import ConfigParser as SafeConfigParser
-except ImportError:
-    from ConfigParser import SafeConfigParser
+import yaml
 
 # Verify Python Version that is running
 try:
@@ -57,7 +54,7 @@ __all__ = [
 
 
 LIB_PATH_ENV_VAR = 'NAPALM_CONF'
-LIB_PATH_DEFAULT = '~/napalm.conf'
+LIB_PATH_DEFAULT = '~/napalm.yml'
 
 
 def get_network_driver(module_name):
@@ -140,24 +137,26 @@ def network_device(named_device, filename=None):
 
     Example (napalm conf file)::
 
-    .. code-block:: python
-        [nxos:n9k1]
-        username: ntc
-        password: pwd123
+    .. code-block:: yaml
+        ---
 
-        [nxos:n9k2]
-        username: ntc
-        password: pwd123
+        n9k1:
+          dev_os: nxos
+          username: ntc
+          password: ntc123
 
-        [ios:csr1]
-        hostname: 10.1.100.10
-        username: ntc
-        password: ntc123
+        csr1:
+          dev_os: ios
+          hostname: 153.92.36.113
+          username: ntc
+          password: ntc123
 
-        [eos:eos-spine1]
-        hostname: 10.1.100.11
-        username: ntc
-        password: ntc123
+        eos-spine1:
+          dev_os: eos
+          hostname: 153.92.39.98
+          username: ntc
+          password: ntc123
+
 
     Example::
 
@@ -172,43 +171,36 @@ def network_device(named_device, filename=None):
 
     """
     config, filename = _get_config_from_file(filename=filename)
-    sections = config.sections()
+    device_kwargs = config.get(named_device)
 
     """
     TODO: NEED TO IMPLEMENT ConfFileNotFoundError
-    if not sections:
-        raise ConfFileNotFoundError(filename)
+    raise ConfFileNotFoundError(filename)
     """
 
-    for section in sections:
-        if ':' in section:
-            device_type, conn_name = section.split(':')
-            if named_device == conn_name:
-                device_kwargs = dict(config.items(section))
-                if 'hostname' not in device_kwargs:
-                    device_kwargs['hostname'] = named_device
+    if device_kwargs:
+        if 'hostname' not in device_kwargs:
+            device_kwargs['hostname'] = named_device
+        device_type = device_kwargs.get('dev_os')
+        driver = get_network_driver(device_type)
+        device = _get_device(driver, device_kwargs)
+        device.open()
+        return device
 
-                driver = get_network_driver(device_type)
-                device = _get_device(driver, device_kwargs)
-                device.open()
-                return device
+    # raise error for device not found
+    """
+    TODO: NEED TO IMPLEMENT
+    raise DeviceNotFoundError(name, filename)
 
-                """
-                TODO: NEED TO IMPLEMENT
-                raise DeviceNotFoundError(name, filename)
-
-                """
+    """
 
 
 def _get_device(driver, device_kwargs):
 
-    core_keys = ['hostname', 'username', 'password', 'timeout']
-    optional_args = dict((k, v) for k, v in device_kwargs.items()
-                         if v is not None and k not in core_keys)
-
     hostname = device_kwargs['hostname']
     un = device_kwargs['username']
     pwd = device_kwargs['password']
+    optional_args = device_kwargs.get('optional_args')
 
     if device_kwargs.get('timeout'):
         timeout = device_kwargs.get('timeout')
@@ -234,7 +226,6 @@ def _get_config_from_file(filename=None):
         else:
             filename = os.path.expanduser(LIB_PATH_DEFAULT)
 
-    config = SafeConfigParser()
-    config.read(filename)
+    config = yaml.load(open(filename))
 
     return config, filename
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -25,8 +25,6 @@ import inspect
 import importlib
 import pkg_resources
 
-import yaml
-
 # Verify Python Version that is running
 try:
     if not(sys.version_info.major == 2 and sys.version_info.minor == 7) and \
@@ -49,12 +47,7 @@ except pkg_resources.DistributionNotFound:
 __all__ = [
     'get_network_driver',  # export the function
     'NetworkDriver',  # also export the base class
-    'network_device'
 ]
-
-
-LIB_PATH_ENV_VAR = 'NAPALM_CONF'
-LIB_PATH_DEFAULT = '~/napalm.yml'
 
 
 def get_network_driver(module_name):
@@ -118,114 +111,3 @@ def get_network_driver(module_name):
         'No class inheriting "napalm_base.base.NetworkDriver" found in "{install_name}".'
         .format(install_name=module_install_name))
 
-
-def network_device(named_device, filename=None):
-    """
-
-    Instantiate and return an instance of a NetworkDriver
-    If no filename is given the environment variable NAPALM_CONF is checked
-    for a path, and finally then ~/napalm.conf is used.
-
-    Arguments:
-        named_device (string): Name of the device as listed in the napalm configuration file.
-        filename (string): (Optional) Path to napalm configuration file that includes
-            the ``named_device`` argument as section header.
-    Raises:
-        DeviceNameNotFoundError: if the named_device is not found in the
-            napalm configuration file.
-        ConfFileNotFoundError: if no napalm configuration file can be found.
-
-    Example (napalm conf file)::
-
-    .. code-block:: yaml
-        ---
-
-        n9k1:
-          dev_os: nxos
-          username: ntc
-          password: ntc123
-
-        csr1:
-          dev_os: ios
-          hostname: 153.92.36.113
-          username: ntc
-          password: ntc123
-
-        eos-spine1:
-          dev_os: eos
-          hostname: 153.92.39.98
-          username: ntc
-          password: ntc123
-
-
-    Example::
-
-
-    .. code-block:: python
-
-        >>> from napalm import network_device
-        >>> device = network_device('csr1')
-        >>> device
-        <napalm_ios.ios.IOSDriver object at 0x7f8e91392210>
-        >>>
-
-    """
-    config, filename = _get_config_from_file(filename=filename)
-    device_kwargs = config.get(named_device)
-
-    """
-    TODO: NEED TO IMPLEMENT ConfFileNotFoundError
-    raise ConfFileNotFoundError(filename)
-    """
-
-    if device_kwargs:
-        if 'hostname' not in device_kwargs:
-            device_kwargs['hostname'] = named_device
-        device_type = device_kwargs.get('dev_os')
-        driver = get_network_driver(device_type)
-        device = _get_device(driver, device_kwargs)
-        device.open()
-        return device
-
-    # raise error for device not found
-    """
-    TODO: NEED TO IMPLEMENT
-    raise DeviceNotFoundError(name, filename)
-
-    """
-
-
-def _get_device(driver, device_kwargs):
-
-    hostname = device_kwargs['hostname']
-    un = device_kwargs['username']
-    pwd = device_kwargs['password']
-    optional_args = device_kwargs.get('optional_args')
-
-    if device_kwargs.get('timeout'):
-        timeout = device_kwargs.get('timeout')
-        if optional_args:
-            device = driver(hostname, un, pwd, timeout=timeout,
-                            optional_args=optional_args)
-        else:
-            device = driver(hostname, un, pwd, timeout=timeout)
-    else:
-        if optional_args:
-            device = driver(hostname, un, pwd,
-                            optional_args=optional_args)
-        else:
-            device = driver(hostname, un, pwd)
-    return device
-
-
-def _get_config_from_file(filename=None):
-
-    if filename is None:
-        if LIB_PATH_ENV_VAR in os.environ:
-            filename = os.path.expanduser(os.environ[LIB_PATH_ENV_VAR])
-        else:
-            filename = os.path.expanduser(LIB_PATH_DEFAULT)
-
-    config = yaml.load(open(filename))
-
-    return config, filename
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -55,8 +55,10 @@ __all__ = [
     'network_device'
 ]
 
+
 LIB_PATH_ENV_VAR = 'NAPALM_CONF'
 LIB_PATH_DEFAULT = '~/napalm.conf'
+
 
 def get_network_driver(module_name):
     """
@@ -118,6 +120,7 @@ def get_network_driver(module_name):
     raise ModuleImportError(
         'No class inheriting "napalm_base.base.NetworkDriver" found in "{install_name}".'
         .format(install_name=module_install_name))
+
 
 def network_device(named_device, filename=None):
     """
@@ -196,6 +199,7 @@ def network_device(named_device, filename=None):
 
                 """
 
+
 def _get_device(driver, device_kwargs):
 
     core_keys = ['hostname', 'username', 'password', 'timeout']
@@ -234,4 +238,3 @@ def _get_config_from_file(filename=None):
     config.read(filename)
 
     return config, filename
-
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -19,10 +19,16 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 # Python std lib
+import os
 import sys
 import inspect
 import importlib
 import pkg_resources
+
+try:
+    from configparser import ConfigParser as SafeConfigParser
+except ImportError:
+    from ConfigParser import SafeConfigParser
 
 # Verify Python Version that is running
 try:
@@ -45,9 +51,12 @@ except pkg_resources.DistributionNotFound:
 
 __all__ = [
     'get_network_driver',  # export the function
-    'NetworkDriver'  # also export the base class
+    'NetworkDriver',  # also export the base class
+    'network_device'
 ]
 
+LIB_PATH_ENV_VAR = 'NAPALM_CONF'
+LIB_PATH_DEFAULT = '~/napalm.conf'
 
 def get_network_driver(module_name):
     """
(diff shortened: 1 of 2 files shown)


mirceaulinic:
> Currently, any parameter that is NOT username, password, hostname, or timeout is considered an optional_arg and will be passed to the NetworkDriver as such.

Why don't you use YAML instead and have the optional_args separately? I find it cleaner that way, e.g.:

```yaml
hostname: lo0
username: beef
password: w00ff
optional_args:
    port: 1234
    config_lock: False
```

jedelman8:
Understood...YAML is an option.  I'm open.  Just easier for many people to not deal with indentation of YAML files :). 

If YAML is preferred, I have no option doing that - it definitely makes the coding side easier.

Whatever is the consensus is, that's fine with me!

ktbyers:
I vote YAML format.

dbarrosop:
I vote for YAML as well

jedelman8:
YAML it is!

jedelman8:
Okay @ktbyers @dbarrosop @mirceaulinic requested changes for the YAML conf file:

Here is an example just tested:

```
---

n9k1:
  dev_os: nxos
  username: ntc
  password: ntc123

csr1:
  dev_os: ios
  hostname: 153.92.36.113
  username: ntc
  password: ntc123

eos-spine1:
  dev_os: eos
  hostname: 153.92.39.98
  username: ntc
  password: ntc123
```

with optional args, it would be:

```
---

n9k1:
  dev_os: nxos
  username: ntc
  password: ntc123

csr1:
  dev_os: ios
  hostname: 153.92.36.113
  username: ntc
  password: ntc123

eos-spine1:
  dev_os: eos
  hostname: 153.92.39.98
  username: ntc
  password: ntc123
  optional_args:
    key: value


```

If you are okay with this, I'll move forward with exceptions, and last, the tests.

ktbyers:
It would be nice to have groups. 

What I did in Netmiko is anything that was a YAML list was a group. A dictionary would be an individual device (like you have here).

```
nxos_devices:
  - n9k1
  - n9k2
  - n9k3
```

Where n9k1, n9k2, n9k3 would be defined in the YAML file.

ktbyers:
I think it would be better to not call open(), if you call open() you limit your options.

What if a person wants to use threads/processes? What if you are returned a list of devices?

I think it is clearer to decouple the inventory (i.e. what is in the YAML file) from what you do with it.

dbarrosop:
I agree, I I'd rather have two functions only:

* `get_group(group_name)` which returns a list of devices that belong to the `group_name`
* `get_device(device_name)` which returns an already instantiated driver but not yet open

So a very simple way of using it would be:

```
my_group = get_group("my_group")

for device in get_group("my_group"):
    with get_device(device) as d:
         d.get_facts()
         ...
```

However, it could also be easily used on more complex scenarios like multithreading as @ktbyers suggested, ie, opening a thread per device returned by `get_group`, passing the device name and running the `get_device` in there.

dbarrosop:
By the way, could you also move this code to an `inventory.py` file or something like that? You can export the method from here but it would be nice to keep it separated so it's a bit cleaner.

ktbyers:
I like @dbarrosop proposal above...

Maybe with an implicit 'all' group i.e. an 'all' group that contains all the devices in the inventory is automatically created.

```
all_devices = get_group("all")
```

dbarrosop:
Yeah, or another method `get_devices` that returns all of them, mostly to avoid hardcoding group names in the logic. Can we also group this functionality inside an `Inventory` class?

jedelman8:
Okay guys, great comments above.  I will put something together and we can review it from there.

jedelman8:
Quick pulse check on preference...what started as an easy task could go in many variations...thoughts on the following for structure of an inventory file:

```
---


n9k1:
  dev_os: nxos
  username: ntc
  password: ntc123
  groups:
    - nexus
    - cisco

csr1:
  dev_os: ios
  hostname: 153.92.36.113
  username: ntc
  password: ntc123
  groups:
    - ios
    - cisco

eos-spine1:
  dev_os: eos
  hostname: 153.92.39.98
  username: ntc
  password: ntc123
  groups: eos

```

**OR**

```

eos:
  - eos-spine1

ios:
  - csr1
nexus:
  - n9k1

cisco:
  children:
    - nexus
    - ios


n9k1:
  dev_os: nxos
  username: ntc
  password: ntc123

csr1:
  dev_os: ios
  hostname: 153.92.36.113
  username: ntc
  password: ntc123

eos-spine1:
  dev_os: eos
  hostname: 153.92.39.98
  username: ntc
  password: ntc123
```

jedelman8:
Here is an example of an input and an output.  From the output, I can return various options like you've said with get_device, get_device, etc. as strings or objects.

The only required key is `group` - notice the last few keys in the yaml file for `group_<name>` .  Any host specific variable will override the group variable.  If no `hostname` is defined, the device name (key) will be used to instantiate a device object.

Are we okay not supporting groups of groups and a single group only for now?


```
---

csr1:
  dev_os: ios
  hostname: 153.92.36.113
  username: ntc
  group: ios

csr2:
  group: ios

eos-spine1:
  dev_os: eos
  hostname: 153.92.39.98
  username: ntc
  password: ntc123
  group: eos


eos-spine2:
  dev_os: eos
  hostname: 153.92.39.99
  username: ntc
  password: ntc123
  group: eos

eos-spine3:
  hostname: 153.92.39.100
  group: eos

eos-spine4:
  group: eos
  password: eos12345

group_eos:
  dev_os: eos
  username: admin
  password: admin

group_ios:
  dev_os: ios
  username: ios123
  password: cisco123
```

output:

```

{
    "eos": {
        "eos-spine1": {
            "dev_os": "eos", 
            "username": "ntc", 
            "group": "eos", 
            "hostname": "153.92.39.98", 
            "password": "ntc123"
        }, 
        "eos-spine3": {
            "dev_os": "eos", 
            "username": "admin", 
            "password": "admin", 
            "group": "eos", 
            "hostname": "153.92.39.100"
        }, 
        "eos-spine2": {
            "dev_os": "eos", 
            "username": "ntc", 
            "group": "eos", 
            "hostname": "153.92.39.99", 
            "password": "ntc123"
        }, 
        "eos-spine4": {
            "dev_os": "eos", 
            "username": "admin", 
            "password": "eos12345", 
            "group": "eos"
        }
    }, 
    "ios": {
        "csr1": {
            "dev_os": "ios", 
            "username": "ntc", 
            "hostname": "153.92.36.113", 
            "group": "ios", 
            "password": "cisco123"
        }, 
        "csr2": {
            "dev_os": "ios", 
            "username": "ios123", 
            "password": "cisco123", 
            "group": "ios"
        }
    }
}

```

ktbyers:
@jedelman8 

Yes, I don't think we really need groups of groups (we can possibly add this longer term).


I think we should start with what methods are we going to have in NAPALM to support this. From earlier discussion:

```
get_group(group_name) which returns a list of devices that belong to group_name
get_device(device_name) which returns an already instantiated driver but not yet open

for device in get_group("my_group"):
    with get_device(device) as d:
         d.get_facts()
         ...
```

And possibly some way of getting all of the groups; all of the devices (in case where you don't know the group_names, device_names).

```
get_groups()           # returns a list of group names as strings
get_devices()          # returns a list of device names as strings
```

Have we decided on this?


A few things I don't really like in the above example you showed:
1. IMO, a device returned should exactly match the key-values for NAPALM i.e. can be fed as **kwargs to NAPALM.
2. A device should be allowed to be multiple groups (not nested groups i.e. but someway of saying device1 is in group1 and in group2).
3. I don't like that we have group link indicated twice in the output:

```
{
    "eos": {                          # group here
        "eos-spine1": {
            "dev_os": "eos", 
            "username": "ntc", 
            "group": "eos",            # group here
            "hostname": "153.92.39.98", 
            "password": "ntc123"
        }, 
```

jedelman8:
Thanks @ktbyers 

I'm going to start with the methods that have been mentioned.  Once I start testing, I may reach out on slack for active feedback.

Regarding you commends here:

A few things I don't really like in the above example you showed:

> IMO, a device returned should exactly match the key-values for NAPALM i.e. can be fed as **kwargs to NAPALM.

Fair enough, and that was the goal.  I will ensure the `group` key is popped and a more complete example with `optional_args`

 > A device should be allowed to be multiple groups (not nested groups i.e. but someway of saying device1 is in group1 and in group2).

I actually had this, but removed it simply because if we use `group_ios` to define params and `group_cisco` as an example, there is no priority for parameters defined.  This would mean, if you put a device in multiple groups, you _should_ define it's parameters (kwargs) at the device level or a default/all group; not technically required, but it would simplify things. Make sense?  

> I don't like that we have group link indicated twice in the output:

Will ensure this is popped this from device object.  As stated, this return object will match key-value pairs for NAPALM.

mirceaulinic:
> Yes, I don't think we really need groups of groups (we can possibly add this longer term).

I think this would be quite useful to have, but, as Kirk mentioned, if hard to implement we should leave it for later.

E.g.:

```yaml
groups:
  juniper_edges:
    devices:
      - device1
      - device2

  cisco_edges:
    devices:
      - device3
      - device4

  edge_routers:
    device_groups:
      - juniper_edges
      - cisco_edges
```

And having this hierarchy of ```groups``` we could also have there some details which will be inherited from the devices underneath as defaults, e.g.:

```yaml
groups:
  juniper_edges:
    username: me
    password: beef

device1:
  group: juniper_edges
  hostname: device1

device2:
  group: juniper_edges
  hostname: device2
  username: its_me_again
  password: strong 
```

Which will expand ```device1``` to:

```json
{
  "device1": {
    "hostname": "device1",
    "username": "me",
    "password": "beef"
  },
  "device2": {
    "hostname": "device2",
    "username": "its_me_again",
    "password": "strong"
  }
}
```

jedelman8:
Can you please look at the YAML file I had previously? The inheritance is there already, but if there are strong opinions on the yaml file structure, please let me know.

mirceaulinic:
> Can you please look at the YAML file I had previously? The inheritance is there already, but if there are strong opinions on the yaml file structure, please let me know.

No, I don't have any strong opinion on this, just ideas :)

Without claiming that's the right path to follow, I don't see where would be the group defaults specified in the YAML from the above? But I do find it handy to have some of the details specified under the group. Imagine you have 100 devices, all using the same username & password: you definitely don't want to specify them for each and every device, but telling what's the group, will take them from there.

mirceaulinic:
Apologies, I just saw it:

```yaml
group_eos:
  dev_os: eos
  username: admin
  password: admin

group_ios:
  dev_os: ios
  username: ios123
  password: cisco123
```

TBH, I don't like that the groups contain their name inside the key, I find it more straight like:

```yaml
groups:
  eos:
    ...etc...
  ios:
    ...etc...
```

But that's probably subjective.

ktbyers:
@jedelman8 What you said earlier makes sense. I was assuming no group parameters so I didn't have some of the issues you were considering.

My vote is we keep the YAML file and file parsing pretty simple...so probably either limit a device to a single group or have no group parameters. I am fine with either.

dbarrosop:
I am fine with not supporting groups of goups, you can always compose them but I think we need to support multiple groups per device. I personally like @mirceaulinic's suggestion. It gets a bit tidier, supports multiple groups per host and it should be doable to nest groups in the future if we want to by adding an extra `groups` keyword alongside the proposed `devices`.

Also, thinking about this, I think it's nice we convert this into a class with very well defined methods that returns something we can consume in a standard way. That way, anybody could implement an inventory for their backend by just overriding the class.

dbarrosop:
In any case, as I mentioned, I am more interested in the Inventory class. If we do a good job with the API that effectively means we can play with different models for the YAML and try to inteagrate with other backends as well.

I am also thinking in creating a `napalm-runner` repo so we can move this there as I have been giving a thought to the idea of creating a simple way of creating a yaml file that defines what to configure where and how to validate the changes and rollback if the validate action fails. Thoughts?

mirceaulinic:
> Also, thinking about this, I think it's nice we convert this into a class with very well defined methods that returns something we can consume in a standard way. That way, anybody could implement an inventory for their backend by just overriding the class.

> I am also thinking in creating a napalm-runner repo so we can move this there as I have been giving a thought to the idea of creating a simple way of creating a yaml file that defines what to configure where and how to validate the changes and rollback if the validate action fails. 

I'm totally into these ideas. Great stuff!

jedelman8:
@mirceaulinic 


> TBH, I don't like that the groups contain their name inside the key (perhaps that's why it wasn't obvious to me where to look for the groups definition);

Actually, this is not the case as `dev_os` does not mean group name

It could be this:

```
group_arista:
  dev_os: eos
  username: admin
  password: admin

group_cisco-ios:
  dev_os: ios
  username: ios123
  password: cisco123
```

Regardless, I'm going to re-read these comments and get something together and will report back.

dbarrosop:
Yeah, no worries, it doesn't have to be perfect. We can iterate on the solution after we have used it a couple of times : )

dbarrosop:
Btw, I am going to create the repo for the "runner". Any suggestion about the name? I thought of "hive" but suggestions are welcome.

mirceaulinic:
I like ```napalm-runner``` more, it would respect the general naming
pattern :)

Le jeu. 15 déc. 2016 00:09, David Barroso <notifications@github.com> a
écrit :

dbarrosop:
Well, it depends on the pattern you are using... I was thinking that `napalm-whatever` denotes a driver for napalm while whatever is just a tool that leverages on napalm.

mirceaulinic:
> Well, it depends on the pattern you are using... I was thinking that napalm-whatever denotes a driver for napalm while whatever is just a tool that leverages on napalm.

Yes, that's true... although we have also ```napalm-utils``` and ```napalm-yang```. It can be anything from my perspective, just that "hive"  I'm not sure it would be quite obvious for the users that it's a runner (I personally would not guess it's that).
And now, to be constructive... I'm afraid that my imagination doesn't want to work today and come with a different solution instead (other than ```napalm-runner``` already exposed) :( 

But I find funny the ```b29``` proposal :) or ```BLU-118```, ``` F-4 Phantom``` 🔥

dbarrosop:
> just that "hive" I'm not sure it would be quite obvious for the users that it's a runner 

I know, that's where marketing comes in xDDD I mean, napalm is not very self-explanatory either xD Anyway, I figured hive as it's a tool to operate multiple devices at once, like a hive mind or it can be short for "High-fiVE" or "Hans the Infallible Virtual Engineer"... I think I need more coffee.

jedelman8:
update this PR along with this one: https://github.com/napalm-automation/napalm/pull/328

New inventory file:

```
---

groups:
  default:
    username: napalm
    password: napalm123
  cisco_ios:
    dev_os: ios
    username: ios-un
    password: ios-pwd
    optional_args:
      foo: bar
  arista:
    dev_os: eos
    username: arista
    password: arista
  dc_edge:
    dev_os: eos
    password: edge456

csr1:
  dev_os: ios
  hostname: 10.1.1.1
  username: ntc
  password: ntc1235
  groups: cisco_ios

csr2:
  groups: cisco_ios

eos-spine1:
  dev_os: eos
  username: a123
  password: e456
  groups:
    - dc_edge
    - arista

eos-spine2:
  hostname: 10.1.1.10
  groups: dc_edge

eos-spine3:
  dev_os: eos
  groups:
    - dc_edge
    - arista


```

If you are using multiple groups for a given device, use device-specific parameters. Otherwise, precedence is device-specific, group-specific, and default.

Here is using the new inventory object:

```
>>> 
>>> from napalm import inventory
>>> 
>>> 
>>> network_inventory = inventory.Inventory('~/napalm2.yml')
>>> 
>>> network_inventory.get_groups()
['arista', 'cisco_ios', 'dc_edge']
>>> 
>>> network_inventory.get_devices()
['eos-spine1', 'eos-spine3', 'csr1', 'csr2', 'eos-spine1', 'eos-spine2', 'eos-spine3']
>>> 
>>> network_inventory.get_devices('dc_edge')
['eos-spine1', 'eos-spine2', 'eos-spine3']
>>> 
>>> 
>>> network_inventory.get_device('csr1')
<napalm_ios.ios.IOSDriver object at 0x7f78cb8d8290>
>>> 
>>> network_inventory.get_group('dc_edge')
[<napalm_eos.eos.EOSDriver object at 0x7f78cb384b50>, <napalm_eos.eos.EOSDriver object at 0x7f78cb384c10>, <napalm_eos.eos.EOSDriver object at 0x7f78cb384ad0>]
>>> 

```

And the full expanded inventory:
```
>>> import json
>>> 
>>> inv = network_inventory.get_inventory()
>>> 
>>> print json.dumps(inv, indent=4)
{
    "arista": {
        "eos-spine1": {
            "dev_os": "eos", 
            "hostname": "eos-spine1", 
            "password": "e456", 
            "username": "a123"
        }, 
        "eos-spine3": {
            "dev_os": "eos", 
            "hostname": "eos-spine3", 
            "password": "arista", 
            "username": "arista"
        }
    }, 
    "cisco_ios": {
        "csr1": {
            "dev_os": "ios", 
            "hostname": "10.1.1.1", 
            "optional_args": {
                "foo": "bar"
            }, 
            "password": "ntc1235", 
            "username": "ntc"
        }, 
        "csr2": {
            "dev_os": "ios", 
            "hostname": "csr2", 
            "optional_args": {
                "foo": "bar"
            }, 
            "password": "ios-pwd", 
            "username": "ios-un"
        }
    }, 
    "dc_edge": {
        "eos-spine1": {
            "dev_os": "eos", 
            "hostname": "eos-spine1", 
            "password": "e456", 
            "username": "a123"
        }, 
        "eos-spine2": {
            "dev_os": "eos", 
            "hostname": "10.1.1.10", 
            "password": "edge456", 
            "username": "napalm"
        }, 
        "eos-spine3": {
            "dev_os": "eos", 
            "hostname": "eos-spine3", 
            "password": "edge456", 
            "username": "napalm"
        }
    }
}
>>> 

```
You can see all usable methods here (also in the `inventory.py` module in the entry point conditional)


    inventory = Inventory('~/napalm2.yml')  # Inventory() would use default settings
    inv = inventory.get_inventory()
    # print json.dumps(inv, indent=4)

    # list of group names
    print inventory.get_groups()

    # list of all devices
    print inventory.get_devices()

    # optionally, pass group name- return list of devices in group
    print inventory.get_devices('dc_edge')

    # return instantiated driver device object for provided device
    print inventory.get_device('csr1')
    print inventory.get_device('eos-spine1')

    # return list of instantiated driver device objects for provided group
    cisco_devices = inventory.get_group('cisco_ios')
    print cisco_devices
    dc_edge_devices = inventory.get_group('dc_edge')
    print dc_edge_devices

ogenstad:
What is the end goal of this? Is it mostly to speed up the creation of napalm device objects, or is it the initial stages of a larger inventory?

Regarding naming for `runner`, another thing would perhaps be that the chosen name should be free at pypi. Perhaps Brigade, as in that which tries to control the fire.

jedelman8:
@ogenstad for me initially, it was just to simplify the creation of napalm device objects.  But at this point, it's becoming a full standalone mini-inventory system (at least the start of one).

Regarding names such as Brigade, that could be a fun exercise.  firetruck, firehose, water, extinguisher, firex (fire extinguiser), i.e. `napalm-firex` :)

dbarrosop:
Brigade it is, I like it and we can't be here discussing about this XD

Can you open the PR against: https://github.com/napalm-automation/brigade

ogenstad:
As a mini-inventory or depending on where you draw the line an inventory. Perhaps there should be a list of feature requests to talk about before starting to write the code for it?

dbarrosop:
I think it's more important we focus on the API and we try it. YAML files and all that is not that relevant as given a decent and simple API you can just rewrite the inventory or add new ones to fit different needs and sources. So I'd recommend focusing on the API and leave the details for later on.

dbarrosop:
Anyway, my plan for this is; we move this to the brigade repo, I have some ideas to build a prototype for the runner. I can mock a prototype and then we can take that as a base to discuss ideas, changes, features, etc...

jedelman8:
@dbarrosop do you want me to close this and issue a PR in the brigade repo providing sample yaml file and inventory script?

dbarrosop:
Yes, let's do that : )

Closing this one myself.

PR #170, opened 2016-12-21
Dbarrosop/framework test bug

dbarrosop:


napalm_base/test/getters.py
```diff
@@ -23,7 +23,7 @@
 from napalm_base.utils.py23_compat import text_type
 
 
-NAPALM_TEST_MOCK = eval(os.getenv('NAPALM_TEST_MOCK', default=True))
+NAPALM_TEST_MOCK = eval(os.getenv('NAPALM_TEST_MOCK', default="1"))
```
dbarrosop: No real reason, I will change it as it's safer.

Issue #114, opened 2016-10-27
Question: default values


mirceaulinic:
I have noticed another inconsistency in terms of default values when some fields cannot be filled with a corresponding value (for various reasons, e.g.: not provided by the vendor, inconsistent value - sometimes , for the same field a stupid vendor returns str or int etc.).

When this happens, we need to return a value though. For string is pretty obvious - returning empty value; for bool I think the right thing is `False` in most of the cases. But how about `int` or `float`? Do you think we should/can keep this homogenous?

Eventually we could define a new module under `napalm-base`, e.g.: `defaults.py` and define a hash such:

``` python
map = {
  str: u'',
  int: -1,
  float: -1.0
}
```

@napalm-automation/council what do you think?

mirceaulinic:
And using the ```py23_compat```https://github.com/ktbyers/napalm-base/blob/ae9b04e6b4d85d56fdd9c9e64b782af44bb16ddd/napalm_base/utils/py23_compat.py  would be even better I guess.

ktbyers:
The default values should be defined in the getter documentation (in napalm-base) and then enforced by the unit tests (if they matter). In other words, where ever we define what the getter should return is where the default values should be specified.

If I was writing a getter, that would be where I would look. In general, I don't really like importing things from napalm-base (for very minor things) as I think it introduces an unnecessary level of indirection. Frequently, this makes something that would be obvious not obvious; then I have to dig into napalm-base to see...oh that is just setting this to a null string.

mirceaulinic:
> The default values should be defined in the getter documentation (in napalm-base) and then enforced by the unit tests (if they matter). In other words, where ever we define what the getter should return is where the default values should be specified.

Totally makes sense, agreed it would be easier.

> If I was writing a getter, that would be where I would look. In general, I don't really like importing things from napalm-base (for very minor things) as I think it introduces an unnecessary level of indirection.

I see your concern, but I don't consider this as a real issue as anyway napalm-base is a dependency.

> Frequently, this makes something that would be obvious not obvious; then I have to dig into napalm-base to see...oh that is just setting this to a null string.

I was thinking something more like:


```python
>>> my_raw_value = 'totally wrong'
>>> napalm_base.helpers.convert(float, my_raw_value)
-1.0
```

And you don't need to look anywhere, the ```convert``` function will try to convert to ```float``` and if it fails, will return the default value for this datatype.

ktbyers:
[...]

PR #171, opened 2016-12-21
WIP: New method OC ACL

bewing:
New method to handle device ACLs

Discussion points:
Do we want to use the OC model as the base (is this PR a non-starter)
How to handle namespacing (EOS supports mac and IP access lists with same name)
Did I handle the flags section clearly, or should we just return a list of text_types with the flag names?
How to handle the port range types OC uses?


itdependsnetworks:
When I think of ACL's/FW rules, I tend to think of them in src -> dst. My 2 cents would be to have them be the dict key, instead of ip proto, eth, and transport. 

Also thinking perhaps support for user and app-id (I know PA, and juniper have that concept.)

dbarrosop:
@bewing I think it makes sense, let me review tomorrow the OC model and get back to you.

@itdependsnetworks this is for ACL, which is not the same as a firewall rule : ) In any case, firewall rules are already implemented here: https://github.com/napalm-automation/napalm-base/blob/develop/napalm_base/base.py#L1453

itdependsnetworks:
@dbarrosop I still stand by the following: 

> When I think of ACL's/FW rules, I tend to think of them in src -> dst. My 2 cents would be to have them be the dict key, instead of ip proto, eth, and transport.

Didn't realize about FW policy, good to know. :) Having looked at ACL's (generally simpler) and a FW policies, I feel one or two of the following is try. FW schema is under-engineered, and/or ACL schema is over engineered.

napalm_base/base.py
```diff
@@ -1499,3 +1499,46 @@ def compliance_report(self, validation_file='validate.yml'):
...
+
+        Args:
+            name(text_type) - Name of ACL to return, default is all
+
+        Returns:
+            A dictionary of ACLs in OC format:
+            * name (dict)
+                * name (text_type)
+                * description (text_type)
+                * acl_entries (dict)
```
bewing: Should this be a list instead of a dict?  If it remains a dict, will sequence_id be the key?

bewing:
Should we omit dictionary keys that are not used in a match?  For example, if we want to match a packet regardless of its DSCP field, should we just leave it out, or have a special value that says "ANY"?  Furthermore, oc-pkt-match-types:port-num-range does support "ANY", so should we specifically set it as "ANY", or omit the key, or leave it up to the driver implementation?

napalm_base/base.py
mirceaulinic: Again I have a personal issue with ACL as Cisco begun implementing them with sequence ID... and others followed. But OC seems to have this as standard (https://github.com/openconfig/public/blob/master/release/models/acl/openconfig-acl.yang#L273) so should we.
To answer your question: yes, we should use it as key - following OC model: https://github.com/openconfig/public/blob/master/release/models/acl/openconfig-acl.yang#L309

napalm_base/base.py
```diff
@@ -1499,3 +1499,46 @@ def compliance_report(self, validation_file='validate.yml'):
...
+                        * source_ip_flow_label (int)
+                        * destination_ip_address (text_type)
+                        * destination_ip_flow_label (int)
+                        * dscp (int)
+                        * protocol (int)
+                        * hop_limit (int)
+                    * transport_fields (dict)
+                        * source_port (text_type)
+                        * destination_port (text_type)
+                        * tcp_flags (dict)
```
mirceaulinic: Here I think it can be a list.

napalm_base/base.py
```diff
@@ -1499,3 +1499,46 @@ def compliance_report(self, validation_file='validate.yml'):
         report file. See https://napalm.readthedocs.io/en/latest/validate.html.
         """
         return validate.compliance_report(self, validation_file=validation_file)
+
+    def get_acls(self, name=''):
```
bewing: I agree on `get_acls_config`, unless we want to leave it open to collecting the state information (per-packet, per-byte statistics, per-interface flags, etc) inside the same function.  I'll wait for @dbarrosop to chime in before changing, though

mirceaulinic: No, the operational data should be collected using a different method, i.e.: ```get_acls_stats``` or smth. See for example ```get_probes_config``` and ```get_probes_results```.

bewing:
> How to handle namespacing (EOS supports mac and IP access lists with same name)

I think that this should be handled on a per-driver basis.  Since OC has indicated that "name" is the key for acl-set, drivers/platforms that have namespace issues MUST prefix OC acl-set names with driver or vendor specific prefixes in order to force uniqueness among IPv4 simple/expanded, IPv6, and MAC ACLs.  Platforms that enforce uniqueness across different classes MAY use a prefix if desired?

mirceaulinic:
> I think that this should be handled on a per-driver basis. Since OC has indicated that "name" is the key for acl-set

Yes, I think so too. Same case with the ACL sequence ID: for example Junos doesn't know about sequence ID, but because the OC model imposed this to be an integer, we'll need to count through them and assign a #. So, yeah, there will be some particular cases handled per driver.

napalm_base/base.py
```diff
@@ -1499,3 +1499,86 @@ def compliance_report(self, validation_file='validate.yml'):
...
+                        * source_ip_flow_label (int)
+                        * destination_ip_address (text_type)
+                        * destination_ip_flow_label (int)
+                        * dscp (int)
+                        * protocol (int)
+                        * hop_limit (int)
+                    * transport_fields (dict)
+                        * source_port (text_type)
+                        * destination_port (text_type)
+                        * tcp_flags (list)
```
mirceaulinic: I was thinking more about just a simple list of flags, without any other nested elements underneath. What do you think?

mirceaulinic:
@bewing if no one else has further comments, I would encourage you to start writing tests. Thanks!

napalm_base/base.py
bewing: How do we want to represent the flag?  String?  Integer referring to a constant?

mirceaulinic: As per https://github.com/openconfig/public/blob/master/release/models/acl/openconfig-packet-match-types.yang#L172, I think it should be a string.

bewing:
So as I'm writing out the tests, I'm back again to how we should handle defaults / missing values / 'any's.

Since I'm trying to use test_model() to test return data, it leans much more towards having empty strings or "any" for wildcards instead of omitting them.  I was playing with testing for existence of keys pointing to dictionaries, so we could skip checking sub-models if they are not present, but I don't think our current test_model() would support checking, for example, the ip_protocol_fields dictionary if DSCP was an optional entry.

Welcome to thoughts, but I'm leaning towards empty strings for wildcards.

mirceaulinic:
Hi @bewing,

> I'm back again to how we should handle defaults / missing values / 'any's.

Regarding defaults, we have a discussion in #114. Looking at this corner case, I think it makes more sense @ktbyers's comment: https://github.com/napalm-automation/napalm-base/issues/114#issuecomment-256686590 - to have the defaults clearly specified in the method documentation.

> Since I'm trying to use test_model() to test return data, it leans much more towards having empty strings or "any" for wildcards instead of omitting them.

I agree.

> I was playing with testing for existence of keys pointing to dictionaries, so we could skip checking sub-models if they are not present, but I don't think our current test_model() would support checking, for example, the ip_protocol_fields dictionary if DSCP was an optional entry.

I'm not sure what do you mean here?
If you refer to the depth of the tests, this depends how and what you are looking to test, e.g.: in your previous proposal you added this line: https://github.com/napalm-automation/napalm-base/blob/develop/napalm_base/test/getters.py#L502 in order to go and test the dictionary under `state`.

> Welcome to thoughts, but I'm leaning towards empty strings for wildcards.

I would be inclined to suggest using `any` instead of empty string for wildcards. But I'm not completely sure about that - perhaps only in time we'll see what's the most suitable.

napalm_base/base.py
```diff
@@ -1499,3 +1499,113 @@ def compliance_report(self, validation_file='validate.yml'):
...
+                    * ip_protocol_fields (dict)
+                        * ip_version (text_type)
+                        * source_ip_address (text_type)
+                        * source_ip_flow_label (int)
+                        * destination_ip_address (text_type)
+                        * destination_ip_flow_label (int)
+                        * dscp (int)
+                        * protocol (int)
+                        * hop_limit (int)
+                    * transport_fields (dict)
```
bewing: Should transport_fields only exist if ip_protocol is 6 or 17?  Should we double-purpose it to match ICMP types and groups?

dbarrosop:
Sorry for the delay with this, flying back and forth and xmas is killing my productivity... I think this one is going to need a bit more care than other methods and we might have to work on the "defaults" we have discussed here:

https://github.com/napalm-automation/napalm-base/issues/114

The reason is that I would like to differentiate in this method what's set, what's not set and what's not supported. For example, if I am getting a L3 ACL I would like to clearly see that the `ethernet_header` is not set.

dbarrosop:
@bewing I am sorry to say you opened a can of worms I have been trying to avoid for a while XDDD

I think we should do something different here. For starters, let's create a branch to work with this. We might need to devise a new method to test OC-compliant models and we might have to tweak the model a bit with nicer defaults.

mirceaulinic:
Looking also at https://github.com/napalm-automation/napalm-base/pull/124, looks like @bewing has a predilection to open cans of worms :)

We could probably release an experimental version of napalm-base and see how it goes - what do yu thing?

dbarrosop:
So I have changed the base branch of this PR and I will be merging it as it is soon. My idea for OC models from now on are as follows:

1. We will have some base code to help dealing with types, optional branches, dependency branches , etc. that YANG defines.
2. OC models will be based off classes instead of dictionaries.

I have been trying to build those dynamically but I am getting to the conclusion it doesn't make sense so let's use this PR as a starting point to play with this.

bewing:
> The reason is that I would like to differentiate in this method what's set, what's not set and what's not supported. For example, if I am getting a L3 ACL I would like to clearly see that the ethernet_header is not set.

Given that the OC implementation did not choose to mark anything in the model as mandatory, perhaps we should be playing looser in base, and giving more latitude to the drivers?  I'd be happy to just leave ethernet_header, transport_fields as optional and test for them with get('ethernet_header', None), and only check their models if they exist.  If they do exist, I still think keeping the defaults inside each section as-is (explicit match or don't-cares like 0.0.0.0/0 for src/dst, -1 for ints, and 'any' for ports, etc.)

dbarrosop:
> perhaps we should be playing looser in base, and giving more latitude to the drivers?

Yes, I completely agree with that. That's why I want to modify how OC methods work and use classes instead of dicts. So the implementer can signal what's not implemented and what's missing due to not being supported by the vendor or what's not yet populated.

If you need this method for production and are in a hurry I suggest you hack it on your own in the way that works best for you as I plan to use this method to try to come up with something for the long term so it might take some time to define this method (2-4 weeks maybe?)

napalm_base/test/getters.py
```diff
@@ -528,5 +528,6 @@ def test_get_acls(self, test_case):
                 if acl_entry.get('transport_fields', None) is not {}:
                     assert helpers.test_model(models.acl_transport_fields,
                                               acl_entry['transport_fields'])
-                assert helpers.test_model(models.acl_action, acl_entry['action'])
+                if acl_entry.get('action', None) is not {}:
```
mirceaulinic: This is not correct: it the `action` key is not set, this test will fail. You could just do `if acl_entry.get('action'):`.

dbarrosop:
Far from usable but it would be nice if you could verify the generated models are correct:

https://github.com/napalm-automation/napalm-yang/pull/4

dbarrosop:
As we progress with OC, I am closing this one. If you think there is something worth discussing, feel free to reopen.

Issue #118, opened 2016-10-29
Enhance `load_template` helper to load config format based on the filename


mirceaulinic:
In https://github.com/napalm-automation/napalm-base/issues/111 we have identified the need to make the helper function ```load_template``` more flexible in terms of config format, depending on the filename.

In order to achieve this, we'll need to include the format type in the filename. Otherwise, by default will be ```text```.
If the template format is different than what the user specifies in the ```optional_args``` as ```config_format```, we'll need to apply a hack to override this format and revert it back after loading the template.

Issue #143, opened 2016-11-18
load_template improvements


ktbyers:
Here are some improvements to load_template() I think we should make.

1. Support 'replace' operation. I would propose that we add a mode='merge' / mode='replace' parameter.

2. Have it behave properly if you specify 'my_template.j2' currently it will look for 'my_template.j2.j2'.

3. Be able to load templates from current working directory (as long as we can make this work in a reasonable way without breaking backwords compatibility).

4. Improve the docs here, http://napalm.readthedocs.io/en/latest/base.html

I think the behavior of some of the parameters needs clarified.

ktbyers:
cc @mirceaulinic 

I can work on these, I just wanted a placeholder so I remember things that I saw.

mirceaulinic:
> Support 'replace' operation. I would propose that we add a mode='merge' / mode='replace' parameter.

I'm into this.

> Have it behave properly if you specify 'my_template.j2' currently it will look for 'my_template.j2.j2'.

Funny one :) Yes, we should definitely avoid this.

> Be able to load templates from current working directory (as long as we can make this work in a reasonable way without breaking backwords compatibility).
> Improve the docs here, http://napalm.readthedocs.io/en/latest/base.html

They all sound good to me @ktbyers! Thanks!

dbarrosop:
> Have it behave properly if you specify 'my_template.j2' currently it will look for 'my_template.j2.j2'.

The idea behind this was that you would specify you want the `ntp` template, so it wasn't really meant to be a file. That's why the `.j2` was added by the method.

> Be able to load templates from current working directory (as long as we can make this work in a reasonable way without breaking backwords compatibility).

If you load the template `blah.j2` in your current folder, that code is gonna be bound to the vendor for which you wrote the template for. If you want to add another vendor later on, your code will have to be adapted. What's wrong with requiring people to put their templates in `./templates/$vendor`? I am trying to see the use case because I think it's effortless and in the long term it might have benefits for the user without any actual drawback.

ktbyers:
The advantage I see is since a lot of platforms copy Cisco for configuration, you could potentially use the same template across multiple platforms (without resorting to symlinks).

The method was confusing when trying to use it as it took several tries to figure out how to get it to work (i.e. the most logical expectation was you could specify 'my_file.j2' and it would look in current working directory). This could be improved, however, by improving the documentation for this method.

dbarrosop:
That the documentation could be improved nobody doubts it :(

PR #173, opened 2016-12-22
`load_template`: config format based on the template extension

mirceaulinic:
For the moment...
Related to #118

### DIFF
--- a/napalm_base/constants.py
+++ b/napalm_base/constants.py
@@ -5,6 +5,11 @@ from __future__ import unicode_literals
 CONFIG_LOCK = True  # must be changed soon!
 TIMEOUT = 60  # seconds
 
+SUPPORTED_CONFIG_FORMATS = [
+    'text'
+]  # all drivers support at least text
+# each driver must override this value
+
 INTERFACE_NULL_SPEED = -1
 
 BGP_NEIGHBOR_NULL_COUNTER = -1
--- a/napalm_base/helpers.py
+++ b/napalm_base/helpers.py
@@ -17,33 +17,62 @@ from netaddr import IPAddress
 
 # local modules
 import napalm_base.exceptions
+import napalm_base.constants as C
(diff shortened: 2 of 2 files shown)


mirceaulinic:
This will allow introducing: https://github.com/napalm-automation/napalm-junos/issues/70

dbarrosop:
Reading #118 I don't think we need this helper code. Let's just ditch completely `cls._config_format` and do something like (junos example):

```
format = filename.split('.')[-1]
if format not in c.SUPPORTED_CONFIG_FORMATS: # Which resolves to ["text", "set", "xml"]
   format = "text"
self.device.load_config(format=format)
```

mirceaulinic:
I agree, just that:

```
self.device.load_config(format=format)
```

the ```format``` arg is currently not accepted for ```load_merge_candidate``` or ```load_replace_candidate```. Or do you mean defining a new public method called ```load_config```?

mirceaulinic:
@dbarrosop changed to handle the configuration format based on the filename.
Also tackles the issue reported by @ktbyers in #143, point 2: now you can call specifying the template name as `my_template.j2` or `my_template.jinja` without appending the `.j2` extension.

dbarrosop:
Sorry I wasn't clear, the `load_config` on my example was the `pyez` one. I didn't recall the exact name so I made it up. So in my version of the implementation, `napalm-base` doesn't require any change. The logic is entirely done on the drivers that have this sort of capability.

mirceaulinic:
> So in my version of the implementation, napalm-base doesn't require any change. The logic is entirely done on the drivers that have this sort of capability.

Oh I see now. But that would require rewriting the whole `load_template` per driver, isn't it?

dbarrosop:
> Oh I see now. But that would require rewriting the whole load_template per driver, isn't it?

Not really, `load_config` relies on the driver's implementation of `load_merge_candidate` so we will have to implement the code there to keep `load_config` generic.

PR #175, opened 2016-12-24
Dbarrosop/validate getter args

dbarrosop:
This adds the possibility to pass arguments to getters via a special keyword `_kwargs`.

I will update the documentation alongside #176


test/unit/validate/test_validate.py
```diff
@@ -81,7 +81,7 @@ def func(**kwargs):
                 with open(filename, 'r') as f:
                     return json.loads(f.read())
             return func
-        if name.startswith("get_"):
+        if name.startswith("get_") or name == "ping":
```
mirceaulinic: Perhaps `if name.startswith("get_") or name in ('ping', 'traceroute')` or `if name.startswith("get_") or name in C.ACTION_TYPE_METHODS` to include also action commands?

dbarrosop: Oh, that's just the mock for the tests, it never pretended to be real or feature-rich. I mean, I could add that but it's going to be dead code ;)

mirceaulinic: Ok, you can ignore my comment in that case :)

PR #181, opened 2016-12-26
Fixed and issue comparing lists of dicts

dbarrosop:
Need to add some unit tests before merging.

### DIFF
--- a/napalm_base/validate.py
+++ b/napalm_base/validate.py
@@ -41,7 +41,8 @@ def _compare_getter_list(src, dst, mode):
         found = False
         for index, dst_element in enumerate(dst):
             intermediate_match = _compare_getter(src_element, dst_element)
-            if intermediate_match:
+            if isinstance(intermediate_match, dict) and intermediate_match["complies"] or \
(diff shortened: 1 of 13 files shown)


XioNoX:
Thanks!
Would it be possible to have a missing/present differences down to the list item and not the full list?

dbarrosop:
@XioNoX I have fixed this issue. I have also added several tests that serve as example to see how this works. Check the use of `get_route_to`.

> Would it be possible to have a missing/present differences down to the list item and not the full list?

Problem is that we are comparing lists of dicts where there is no unique "key" I can use to determine that they are the same object and, thus, show the diff. For example:

https://github.com/napalm-automation/napalm-base/pull/181/files#diff-8a0539f4f2d874e6f8149c720c55a1fdR2

Should I compare the route in the `validate.yml` against the first or the second route? If you have any idea on how to do this in a generic way, let me know.

XioNoX:
@dbarrosop Indeed, I looked at it and can't think of an easy solution.

Issue #183, opened 2017-01-04
Adding functions for get_eigrp_config() and get_eigrp_neighbors()


WilliamMarti:
So today there are the following methods in napalm-base

`get_bgp_neighbors()`
`get_bgp_config()`

In the network I support we are heavy EIGRP users and could use similar functionality.  I think this comes back to a larger discussion from slack and [here](https://github.com/napalm-automation/napalm-base/issues/182), talking about vendor specific protocols.  So my proposal is to add:

`get_eigrp_neighbors()`
`get_eigrp_config()`

Realistically these would only ever be implemented in the Cisco napalm flavors.  

Looking at the `get_route_to` function, it should already support EIGRP routes.

dbarrosop:
Yeah, I think those methods are fine. Is this something you are looking for to implement yourself?

mirceaulinic:
Hi @WilliamMarti - are you still interested in this?

WilliamMarti:
Apologies, totally fell off my radar.  

Yes I would like to implement this.  I will get working on the PR for napalm-base now.

mirceaulinic:
@WilliamMarti no problem at all - just checking

dbarrosop:
This looks good in principle, I would like to see an initial implementation though before approving and merging.

PR #197, opened 2017-02-01
Addition of get_eigrp_neighbors()

WilliamMarti:
Discussion at https://github.com/napalm-automation/napalm-base/issues/183.

Starting with just the get_eigrp_neighbors() method for now to get my feet wet with the code base before I get too deep.


napalm_base/test/base.py
```diff
@@ -569,3 +569,13 @@ def test_get_network_instances(self):
...
+
+    def test_get_eigrp_neighbors(self):
+        """Test get_eigrp_neighbors method."""
+        try:
+            get_eigrp_neighbors = self.device.get_eigrp_neighbors()
+        except NotImplementedError:
+            raise SkipTest()
+
+        assert isinstance(get_eigrp_neighbors, dict)
+        assert self._test_model(models.eigrp_config_neighbors, get_eigrp_neighbors)
```
mirceaulinic: newline please

WilliamMarti:
All doable, based on the comments from @dbarrosop in https://github.com/napalm-automation/napalm-base/issues/183 I will wait until I have an actual implementation before my next PR.

Thanks guys

mirceaulinic:
Hi @WilliamMarti - how is this going? Just checking if you need any help with the implementation.

icmp-echo:
it would be helpful if we also can get something like get_vlan and get_interfaces_descriptions because that would help us lot in Network automation project, i am not a developer i am new to this but i am trying to wrap my head around how to write those modules so I can also contribute towards that. Any help is appreciated. I wanna work on this if somebody can point me right direction.

dbarrosop:
@icmp-echo you have some info here: https://napalm.readthedocs.io/en/latest/contributing/index.html#proposing-a-new-method

Btw, `get_interfaces_descriptions ` shouldn't be necessary as `get_interfaces` already gives you the interface description.

icmp-echo:
@dbarrosop Thank you for responding, I really appreciate it.
About get_interface You are right sir,
 it does pull interfaces description now question becomes can we use grain or pass an argument to just return interface description for specific port  like this:

salt -G 'os:junos and interface:fe-0/0/5' grains.get description 

we dont have a grain for description yet, but i envision to pull that off same with the vlans as well, once we have get_vlan it would be easy to write the grains on top of that.

but this a good start. Thank you.

dbarrosop:
`salt -G 'os:junos and interface:fe-0/0/5' grains.get description`

That looks like a salt thing, I don't think we need a method for that. @mirceaulinic can you comment on this? You are the resident salt expert here ;)

mirceaulinic:
`salt -G 'os:junos and interface:fe-0/0/5'` selects those devices running Junos and having the interface `fe-0/0/5`. The grains are used to select minions, not a particular feature / subset / subsystem.

I see what you are trying to achieve @icmp-echo. Searching for a certain for a particular detail indeed is very useful in production networks especially when it grows that much that you can't keep in mind all the details.
For this, you may find this Salt runner useful: https://docs.saltstack.com/en/develop/ref/runners/all/salt.runners.net.html -- will be added in the next release, Nitrogen (later this month). To see how this runner works, see this video: https://www.youtube.com/watch?v=AqBk5fM7qZ0. In the first example it does returns all interfaces having the pattern "cogent" in their description. Similarly you can search for mac addresses, ip addresses, interfaces, etc. In your case would probably be `salt-run net.find fe-0/0/5`.

As an aside, check this out too: https://docs.saltstack.com/en/develop/ref/runners/all/salt.runners.bgp.html.

You can also use the grains to achieve this, but I would not recommend. Salt impresses with its speed and caching the interfaces under grains would slow down a bit the matching if you have a device with many interfaces. Grains should have a very simple structure.

But if you really want, you can, by writing a custom grains modules in your envirnment: https://docs.saltstack.com/en/develop/topics/grains/index.html#writing-grains
See below an example:

```python
def interfaces_grain(proxy):
    if proxy['napalm.initialized']:
        intrf = proxy['napalm.call']('get_interfaces')
        if intrf:
            return {'interfaces_details': intrf.get('out', {})}
 ```

And then you can select your devices like: `salt -G 'os:junos and interfaces_details:fe-0/0/5:description:*pattern*`

icmp-echo:
@mirceaulinic this is awesome feature, and more I learn about salt more I fall in love with its capacity and power it posses, thank you for your awesome response, It a lot for me to process and think in your one response, but again thanks alot. 
@dbarrosop I am gonna open a new issue but i dont know how to move existing replies & comments to a new issue, but its a good idea. Thank you.

icmp-echo:
new issue #260 was opened for follow up @dbarrosop . Thanks.

mirceaulinic:
Hi @WilliamMarti - continuing from https://github.com/napalm-automation/napalm-base/pull/197#issuecomment-276672302: are you still working on this?

Issue #180, opened 2016-12-26
compliance_report doesn't support several ping targets


XioNoX:
Currently, if the validate.yml file contains several ping statements

```
ping:
  _kwargs:
      destination: 8.8.8.8
      source: 192.168.1.1
  success:
      packet_loss: 0
  _mode: strict

ping:
  _kwargs:
      destination: 4.4.2.2
      source: 192.168.1.1
  success:
      packet_loss: 0
  _mode: strict
```
The end result will only consider one of the two:
```
            "ping": {
                "complies": true, 
                "extra": [], 
                "missing": [], 
                "present": {
                    "success": {
                        "complies": true, 
                        "nested": true
                    }
                }
            }
```

It would be useful to be able to test many different ping tasks for a given host. The same might apply to traceroute.

dbarrosop:
@napalm-automation/council

I see two options here:

1. Replace the key with a name. Something like:

```
test_pinging_google_dns:
	ping:
	  _kwargs:
	      destination: 8.8.8.8
	      source: 192.168.1.1
	  success:
	      packet_loss: 0
	  _mode: strict

test_pinging_other_google_dns:
	ping:
	  _kwargs:
	      destination: 4.4.2.2
	      source: 192.168.1.1
	  success:
	      packet_loss: 0
	  _mode: strict
```

2. composing the keys to be able to add a key (separting command and description with two `_`:

```
ping__google_dns:
  _kwargs:
      destination: 8.8.8.8
      source: 192.168.1.1
  success:
      packet_loss: 0
  _mode: strict

ping__other_google_dns:
  _kwargs:
      destination: 4.4.2.2
      source: 192.168.1.1
  success:
      packet_loss: 0
  _mode: strict
```

I personally prefer the second option but I am open to suggestions/opinions.

mirceaulinic:
@dbarrosop I do find the first option _much_ cleaner. I personally want to avoid hacks as in the second one.

dbarrosop:
There are actually two reasons I like the second method:

1. Most methods don't require a name as it doesn't make sense to execute two time `get_facts` or `get_bgp_neighbors`. Name only makes sense on methods that uses arguments.
2. Using `ping__blah` serves as a key on the report as well.

Maybe we can do someting like:

```
get_facts:
   os_version: blah

ping:
  _name: google dns
  _kwargs:
      destination: 8.8.8.8
      source: 192.168.1.1
  success:
      packet_loss: 0
  _mode: strict

[...]

PR #202, opened 2017-02-07
Fix for multiple getters

XioNoX:
Fix for issue reported in: https://github.com/napalm-automation/napalm-base/issues/180#issuecomment-277608980
After getting a suggestion of how to fix it from @dbarrosop 

Tested with validation files consisting of either lists (new format) and dicts (old format)

### DIFF
--- a/napalm_base/validate.py
+++ b/napalm_base/validate.py
@@ -131,19 +131,22 @@ def compliance_report(cls, validation_file=None):
     report = {}
     validation_source = _get_validation_file(validation_file)
 
-    for getter, expected_results in validation_source.items():
-        if getter == "get_config":
-            # TBD
-            pass
-        else:
-            key = expected_results.pop("_name", "") or getter
+    if isinstance(validation_source, dict):
+        validation_source = [validation_source]
+    for validation_check in validation_source:
+        for getter, expected_results in validation_check.items():
+            if getter == "get_config":
+                # TBD
+                pass
+            else:
+                key = expected_results.pop("_name", "") or getter
 
-            try:
-                kwargs = expected_results.pop('_kwargs', {})
-                actual_results = getattr(cls, getter)(**kwargs)
-                report[key] = _compare_getter(expected_results, actual_results)
-            except NotImplementedError:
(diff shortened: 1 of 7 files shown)


napalm_base/validate.py
```diff
@@ -131,19 +131,22 @@ def compliance_report(cls, validation_file=None):
...
-        else:
-            key = expected_results.pop("_name", "") or getter
-
-            try:
-                kwargs = expected_results.pop('_kwargs', {})
-                actual_results = getattr(cls, getter)(**kwargs)
-                report[key] = _compare_getter(expected_results, actual_results)
-            except NotImplementedError:
-                report[key] = {"skipped": True, "reason": "NotImplemented"}
+    if isinstance(validation_source, dict):
```
dbarrosop: Don't do this, we should support only one method.

XioNoX: Maybe use both methods during a transition period? The issue here is that the necessary "new" method would breaks if people are using files with the "old" one.
I'm happy to remove that part if you think it's not worth it.

dbarrosop: I think it's ok to break it as supporting both might have interesting and unexpected implications on the recursive methods.

XioNoX: No problem! patch updated!

dbarrosop:
You will have to fix the CI :)

XioNoX:
CI fixed!
I also added the following tests for _name values:
```yaml
- ping:
    _name: ping1
    _kwargs:
        destination: 8.8.8.8
        source: 192.168.1.1
    success:
        packet_loss: 0
    _mode: strict

- ping:
    _name: ping2
    _kwargs:
        destination: 8.8.8.8
        source: 192.168.1.1
    success:
        packet_loss: 0
    _mode: strict
```
@dbarrosop anything left?

dbarrosop:
Ok, now there are conflicts here... I guess that we can close it anyway as they changes here were also on #203?

XioNoX:
Correct, #203 includes the changes from this PR.

dbarrosop:
Ok, closing this one then. Thanks again for the PR : )

PR #206, opened 2017-02-11
interfaces_ip primary IP check

itdependsnetworks:
For IPv4 addresses, look to add check for primary IP. This method should be backwards compatible with current implementation.

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -796,6 +796,7 @@ class NetworkDriver(object):
         addresses as keys.
         Each IP Address dictionary has the following keys:
             * prefix_length (int)
+            * primary_ip (boolean)
 
         Example::
 
@@ -803,7 +804,11 @@ class NetworkDriver(object):
                 u'FastEthernet8': {
                     u'ipv4': {
                         u'10.66.43.169': {
-                            'prefix_length': 22
+                            'prefix_length': 22,
+                            'primary_ip': True
+                        },
+                        u'10.66.50.1': {
+                            'prefix_length': 24
                         }
                     }
                 },
@@ -831,7 +836,8 @@ class NetworkDriver(object):
                 u'Tunnel0': {
                     u'ipv4': {
                         u'10.63.100.9': {
-                            'prefix_length': 24
+                            'prefix_length': 24,
+                            'primary_ip': True
                         }
                     }
                 }


napalm_base/base.py
```diff
@@ -796,14 +796,19 @@ def get_interfaces_ip(self):
...
             {
                 u'FastEthernet8': {
                     u'ipv4': {
                         u'10.66.43.169': {
-                            'prefix_length': 22
+                            'prefix_length': 22,
+                            'primary_ip': True
+                        },
+                        u'10.66.50.1': {
+                            'prefix_length': 24
```
mirceaulinic: Add the primary key here as well, please

mirceaulinic:
As this is a API change, we'll need to apply the changes for the following drivers:

- eos
- ios
- iosxr
- junos
- nxos
- ros
- vyos

So it might take a while to introduce this.
@itdependsnetworks if you can bring some help with this, it would be much appreciated.

dbarrosop:
Could you create a label and tag all the issues/PRs related to this with it so it's easier to track with zenhub?

Thanks!

mirceaulinic:
Yeah.. lemme install it first and play :)

mirceaulinic:
Hi @itdependsnetworks - would you be able to provide an implementation for one or more drivers? Here you can see the issues linked to the parent: https://github.com/napalm-automation/napalm-base/issues/207. Thanks!

itdependsnetworks:
Apologies for not getting back. Yea, Should be able to on ios, nxos, junos, and eos. Give a week or two.

narJH27:
@itdependsnetworks @mirceaulinic Any update on this enhancement wrt updating the individual drivers?

itdependsnetworks:
Been crazy busy, will try to get to this on the weekend... But I have said that before :(

narJH27:
@itdependsnetworks Sounds good! In case you are unable to get some of it done by early next week, let me know and maybe I can help out.

narJH27:
@mirceaulinic @itdependsnetworks I have implemented the changes for napalm_base, eos, and iosxr. Pending merge to base, for the other PR's to pass the checks.

napalm_base/base.py
```diff
@@ -808,28 +808,34 @@ def get_interfaces_ip(self):
...
                     u'ipv6': {
                         u'1::1': {
-                            'prefix_length': 64
+                            'prefix_length': 64,
+                            'primary_ip': False
                         },
                         u'2001:DB8:1::1': {
-                            'prefix_length': 64
+                            'prefix_length': 64,
+                            'primary_ip': True
```
dbarrosop: There is no concept of primary or secondary ip on IPv6.

itdependsnetworks: Interesting, so even when a router needs to source an IP, how does it choose which one?

Does it make sense not to include on IPv6 at all?

dbarrosop: There is a whole RFC for that (surprise!): https://tools.ietf.org/html/rfc6724
Actually, IPv4 doesn't have the concept at all either. AFAICT only Cisco has the "secondary" keyword. Juniper certainly doesn't.

> Does it make sense not to include on IPv6 at all?

You are probably right. You will have to slightly adapt the test so it uses two different models; one for IPv4 and one for IPv6.

PR #219, opened 2017-03-14
Add get_static_rp_config method and test

bewing:
New method to retrieve PIM static RP configuration.

This is built around the EOS CLI parser, so I'm open to suggestions from other drivers on changing the return/output.

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -1492,6 +1492,26 @@ class NetworkDriver(object):
         """
         raise NotImplementedError
 
+    def get_static_rp_config(self):
+        """
+        Returns a dictionary of VRFs and their configured static PIM RPs, where unset keywords
+        are type None.
+
+        Example:
+
+        {
+            u'default': [{
+                u'acl': None,
+                u'bsr_override': None,
+                u'hashmask': None,
+                u'mcast_subnet': '224.0.0.0/4',
+                u'priority': None,
+                u'rp_addr': '192.0.2.1'
+            }]
+        }
+        """
+        raise NotImplementedError
+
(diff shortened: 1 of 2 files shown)


dbarrosop:
Can we base this on the IETF model instead? If this can hold for 2-3 weeks we might be able to even launch this as part of the OpenConfig framework.

bewing:
I don't have an issue with holding this.

bewing:
For reference, the IETF model for static RPs:
```
{
    "pim": {
        "address-family": {
            "ipv4": {
                "address-family": "ipv4",
                "rp": {
                    "static-rp": {
                        "ipv4-rp": {
                            "192.0.2.1": {
                                "ipv4-address": "192.0.2.1",
                                "pim-sm": {
                                    "policy-name": "v4-sm-acl",
                                    "override": False,
                                    "priority": 0, # EOS ONLY
                                    "hash-mask-length": 30 # EOS ONLY
                                },
                                "pim-bidir": {
                                    "policy-name": "v4-bidir-acl",
                                    "override": False,
                                }
                            }
                        }
                    }
                }
            },
            "ipv6": {
                "address-family": "ipv6",
                "rp": {
                    "static-rp": {
                        "ipv6-rp": {
                            "2001:DB8::1": {
                                "ipv6-address": "2001:DB8::1",
                                "pim-sm": {
                                    "policy-name": "v6-sm-acl",
                                    "override": False,
                                },
                                "pim-bidir": {
                                    "policy-name": "v6-bidir-acl",
                                    "override": False,
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}
```

mirceaulinic:
@bewing - what's the status here?

PR #233, opened 2017-04-04
Test case for static routes (v4+v6)

manuel-domke:
hi,

I need test cases for static ipv4/ipv6 routes for my changes made in napalm-pyiosxr.

### DIFF
--- a/napalm_base/test/getters.py
+++ b/napalm_base/test/getters.py
@@ -366,6 +366,21 @@ class BaseTestGetters(object):
         return get_route_to
 
     @wrap_test_cases
+    def test_get_route_to_v6(self, test_case):
+        """Test get_route_to_v6."""
+        destination = 'dead:beef:210:210::53/64'
+        protocol = 'static'
+        get_route_to = self.device.get_route_to(destination=destination, protocol=protocol)
+
+        assert len(get_route_to) > 0
+
+        for prefix, routes in get_route_to.items():
+            for route in routes:
+                assert helpers.test_model(models.route, route)
+
+        return get_route_to
+
+    @wrap_test_cases
     def test_get_snmp_information(self, test_case):
         """Test get_snmp_information."""
         get_snmp_information = self.device.get_snmp_information()
--- a/napalm_base/test/getters.py
+++ b/napalm_base/test/getters.py
@@ -366,8 +366,23 @@ class BaseTestGetters(object):
         return get_route_to
 
     @wrap_test_cases
-    def test_get_route_to_v6(self, test_case):
-        """Test get_route_to_v6."""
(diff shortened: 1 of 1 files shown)


dbarrosop:
Ok, adding these tests basically break CIs for other drivers so we will have to be careful with this one.

Drivers we have to fix:

1. EOS
2. IOSXR
3. Junos
4. Panos

Shouldn't be terrible.

However, I was wondering if we should just remove the "protocol" from `test_get_route_to` so we can control with the test cases (default and the rest) the data we get and parse instead.

@manuel-domke @mirceaulinic any thoughts?

manuel-domke:
I was wondering on how many of the other implementations I'll break. Glad its just three :)
And - yes. Lets remove the protocol for "default" test cases.

dbarrosop:
See: https://github.com/napalm-automation/napalm-base/issues/238

mirceaulinic:
I need to think more about it, but for the moment I'm not sure I can agree with this: do we want to have more bespoken methods just for the sake of testing and argument? If the protocol is too problematic, we could just ignore it, as we do for other methods, e.g.: `get_bgp_neighbors_detail` or others having optional arguments. What would we do it we'd need to test the argument `neighbor` for `get_bgp_neighbors_detail`? Would we implement `get_bgp_neighbors_detail_neighbor`?

dbarrosop:
> remove an useful functionality

Nobody wants to remove anything, just move it around.

> why not just implement ping_vrf, ping_ttl

Easy to answer that and I will just refer to the code:

https://github.com/napalm-automation/napalm-eos/blob/develop/napalm_eos/eos.py#L1054

Protocol is a switch to go to a branch of the code that contains ~100 lines of code. Adding a new protocol implies branching again and executing 100 other different lines of code.

In any case, please, read https://github.com/napalm-automation/napalm-base/issues/238 (point 1). Nobody mentioned removing anything, jut moving `get_route_to` to `napalm-base` and have it call different methods for each individual protocol. Something like:

```
# This is not working code, just for reference
def get_route_to(self, protocol='', ...):
    if protocol in ["direct", "bgp", "ospf", "static"...]:
        method = getattr(self, "get_route_to_{}".format(protocol))
        return method(...)
    elif protocol == "":
       # code to call all of getters and ignore notimplementederrors
   else:
      # fail as protocol is incorrect
```

And then each driver just has to implement `get_route_to_static`, `get_route_to_bgp`, etc. which serves as implementation and documentation. You can refer to the docs and see that `get_route_ospf` is not implemented and thus it won't work. Right now it's just ignored as you suggested.

dbarrosop:
Response here: https://github.com/napalm-automation/napalm-base/issues/238#issuecomment-296098915

PR #235, opened 2017-04-08
Check if connection is alive before closing

dbarrosop:
Fixes https://github.com/napalm-automation/napalm-ios/issues/135

@gjim83 @mirceaulinic comments?

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -66,11 +66,10 @@ class NetworkDriver(object):
         is released (unlocked).
         """
         try:
-            self.close()
-        except OSError as e:
-            # Ignore if socket was already closed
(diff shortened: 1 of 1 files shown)


gjim83:
This is what I thought at some point while discussing https://github.com/napalm-automation/napalm-base/commit/842a801aba325b0bd85772cb9a5fd2e87ecfe6f2 but then decided back on ignoring the exception. That was because I wasn't sure I was aware of all the actions that all drivers take when closing, that wouldn't be tied to closing the socket (when applicable).
However, I'd say this works safely, more so when taking into account this comment here https://github.com/paramiko/paramiko/blob/master/paramiko/channel.py#L1133 :)

PR #239, opened 2017-04-19
Add support for transition of signatures (to ease migration of methods)

ktbyers:
Adds a new data structure TRANSITION_SIGNATURES that allows for easier migration of methods from old to new state.

napalm-base would be pushed first with the new signature.
Then all napalm-drivers would be pushed with new signature.
Then TRANSITION_SIGNATURES would be commented out (i.e. set to blank list until next migration of signatures occur).

### DIFF
--- a/napalm_base/test/getters.py
+++ b/napalm_base/test/getters.py
@@ -122,6 +122,24 @@ class BaseTestGetters(object):
 
     def test_method_signatures(self):
         """Test that all methods have the same signature."""
+
+        # Method signatures that are migrating from old to new state (temporary state)
+        transition = True
+        if transition:
+            from inspect import ArgSpec
+        # napalm-base must match new; napalm-driver must match old (if not identical to base)
+        TRANSITION_SIGNATURES = {
+            'commit_config': {
+                'old': ArgSpec(args=['self'], varargs=None, keywords=None, defaults=None),
+                'new': ArgSpec(args=['self', 'confirmed'], varargs=None,
+                               keywords=None, defaults=(0,)),
+            },
+            'commit_confirm': {
+                'old': 'Method does not exist in napalm_base',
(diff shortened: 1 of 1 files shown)


ktbyers:
This needs to be pushed with:

https://github.com/napalm-automation/napalm-base/pull/213

I have tested with napalm-ios. 

Once pushed, I will further regression test against this PR.

https://github.com/napalm-automation/napalm-ios/pull/124

napalm_base/test/getters.py
```diff
@@ -122,6 +122,24 @@ class BaseTestGetters(object):
 
     def test_method_signatures(self):
         """Test that all methods have the same signature."""
+
+        # Method signatures that are migrating from old to new state (temporary state)
+        transition = True
+        if transition:
```
mirceaulinic: I'm not sure I understand this :)

ktbyers: Yes...that isn't very good...let me cleanup.

dbarrosop:
Let's discuss it in the meeting next week. This has big implication so I'd like to discuss it and make sure everybody is on the same page before moving forward. My fear is that this basically opens the door to non-compatible code. You could install `napalm` and have two drivers with two different method signatures and write incompatible code. A solution could be to add `*args, **kwargs` to all signatures.

ktbyers:
@dbarrosop @mirceaulinic Can we revisit this as a transition solution for migrating the API? Without something like this progress on these API changes grinds to a halt.

dbarrosop:
Yes, we should proceed with this. I have one comment though. Could we have a yaml file instead in the root of the repo where we put the signatures instead of buried in the code? The reason is to have it on a more visible place so we don't forget about them. It would be great also if we could trigger a huge warning during the tests just to annoy us so we don't grow lazy because of this :P

PR #246, opened 2017-05-27
Update base helpers to include as_number and add a version to ip

ktbyers:
Also remove some obsolete PY3 transition code.


dbarrosop:
This is great, so much cleaner! :D

napalm_base/helpers.py
```diff
@@ -239,4 +242,16 @@ def ip(addr):
...
         u'2001:db8:85a3::8a2e:370:7334'
     """
-    return py23_compat.text_type(IPAddress(addr))
+    addr_obj = IPAddress(addr)
+    if version and addr_obj.version != version:
+        raise ValueError("{} is not an ipv{} address".format(addr, version))
+    return py23_compat.text_type(addr_obj)
+
+
+def as_number(as_number):
```
mirceaulinic: Hrm, I'm surprised the linter didn't complain about this (the name of the variable and the name of the function are the same).

ktbyers: Yeah...that was dumb on my part as was not checking for the integer ASN case. 

It is probably valid as one variable is in the global namespace, the other variable is local to the function. Still not a good idea...

Issue #218, opened 2017-03-06
Better exceptions


mirceaulinic:
We will need to create few more exceptions, to avoid as much as possible propagating exceptions from the underlying libraries.

Some examples:

- LockError
- UnlockError
- CommitError
- ConnectAuthError
- ConnectTimeoutError

mirceaulinic:
Solved under https://github.com/napalm-automation/napalm-base/pull/248, but leaving this open for future ideas.

PR #248, opened 2017-05-30
More exceptions

mirceaulinic:
Define new exception `ConnectionClosedException` as per https://github.com/napalm-automation/napalm-base/pull/247

Add the exceptions proposed in https://github.com/napalm-automation/napalm-base/issues/218

### DIFF
--- a/napalm_base/exceptions.py
+++ b/napalm_base/exceptions.py
@@ -24,7 +24,6 @@ class ModuleImportError(Exception):
 class ConnectionException(Exception):
     '''
     Unable to connect to the network device.
-    Raised by the ``open`` method.
     '''
     pass
 
@@ -39,7 +38,10 @@ class ConnectAuthError(ConnectionException):
 
 class ConnectTimeoutError(ConnectionException):
     '''
-    Connection to the network device takes too long.
+    Exception raised when the connection to the
+    network device takes too long.
+    This may be avoided by adjusting the `timeout`
+    argument.
     '''
     pass
 
@@ -48,8 +50,11 @@ class ConnectionClosedException(ConnectionException):
     '''
     The network device closed the connection.
     Raised whenever we try to execute a certain
-    function, but we detect that the device dropped
-    the connection.
+    function, but we detect that the connection
+    is not usable anymore. This can happen for
+    various reasons: the network device terminates the
+    session or it is dropped by a firewall or
+    the server.
     '''
     pass
 
--- a/napalm_base/exceptions.py
+++ b/napalm_base/exceptions.py
@@ -22,6 +22,20 @@ class ModuleImportError(Exception):
 
 
 class ConnectionException(Exception):
+    '''
+    Unable to connect to the network device.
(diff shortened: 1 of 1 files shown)


napalm_base/exceptions.py
```diff
@@ -22,6 +22,35 @@ class ModuleImportError(Exception):
 
 
 class ConnectionException(Exception):
+    '''
+    Unable to connect to the network device.
+    Raised by the ``open`` method.
```
dbarrosop: This last sentence is not accurate. For two reasons;

1. childs of this class will be thrown on other places, which means you can actually catch this exception in other methods.
2. It kind of implies this is the only exception that is going to be thrown by the open method while the truth is that it would be nice if its childs were raised instead.

napalm_base/exceptions.py
```diff
@@ -22,6 +22,35 @@ class ModuleImportError(Exception):
...
+    '''
+    Unable to connect to the network device
+    due to invalid credentials.
+    '''
+    pass
+
+
+class ConnectTimeoutError(ConnectionException):
+    '''
+    Connection to the network device takes too long.
```
dbarrosop: Maybe mention timeout value can be tweaked with the timeout argument. Mostly to indicate there is a correlation.

napalm_base/exceptions.py
```diff
@@ -22,6 +22,35 @@ class ModuleImportError(Exception):
...
+    '''
+    pass
+
+
+class ConnectionClosedException(ConnectionException):
+    '''
+    The network device closed the connection.
```
dbarrosop: I would reword this and use "dropped' instead unless we are expecting to raise this exception if the server ends the TCP socket. My guess is we will be raising this if the connection is silently dropped by a firewall or the server.

mirceaulinic:
@dbarrosop I've pushed some changes: https://github.com/napalm-automation/napalm-base/pull/248/commits/b3b5c9be7982169f43968fa5af9291e8af1c2807, feel free to suggest more improvements if anything still doesn't sound good.

ktbyers:
In SSH you can also have the SSH channel closed by remote end (currently raises an EOFError in Netmiko). In this case the TCP connection is still open, but you are unable to communicate.

I think this should probably fall into this `ConnectionClosedException` as you would need to treat it the same way in practice (i.e. re-open a new connection).

mirceaulinic:
Yes, that was my understanding too @ktbyers. Would you suggest including this explanation in the exception docsting?

ktbyers:
I think we might want to include that in the docstring (i.e. 'or SSH channel closing').

dbarrosop:
We don't really need super explicit. Most of the time people will just want to know the connection was dropped due to inactivity (regardless of who or how) and just try to reconnect. My point is that a single ConnectionDroppedException and then different messages for logging purposes should probably be enough. I can't imagine someone treating a timeout because a firewall dropped a flow being treated different than ssh closing the channel.

PR #255, opened 2017-06-02
Fix the as_number helper

mirceaulinic:
- confusing variable and function have the same name
- fails when the value is already an integer


dbarrosop:
LGTM but I think it would be great if you could add some unit tests like these: https://github.com/napalm-automation/napalm-base/blob/develop/test/unit/TestHelpers.py#L320

mirceaulinic:
Right, forgot about tests...

PR #259, opened 2017-06-05
Define generic napalm metaclass to catch exceptions

mirceaulinic:
And raise a napalm exception instead.
As discussed under https://github.com/napalm-automation/napalm-junos/pull/161

Log trace:

```
[ERROR   ] Raised jnpr.junos.exception.ConnectClosedError
  File "/home/admin/napalm-base/napalm_base/base.py", line 50, in fun
    return meth(*args, **kwargs)
  File "/home/admin/napalm-junos/napalm_junos/junos.py", line 1069, in get_arp_table
    arp_table_raw.get()
  File "/usr/local/lib/python2.7/dist-packages/jnpr/junos/factory/optable.py", line 64, in get
    self.xml = getattr(self.RPC, self.GET_RPC)(**rpc_args)
  File "/usr/local/lib/python2.7/dist-packages/jnpr/junos/rpcmeta.py", line 336, in _exec_rpc
    return self._junos.execute(rpc, **dec_args)
  File "/usr/local/lib/python2.7/dist-packages/jnpr/junos/decorators.py", line 63, in wrapper
    result = function(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/jnpr/junos/decorators.py", line 31, in wrapper
    return function(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/jnpr/junos/device.py", line 695, in execute
    raise EzErrors.ConnectClosedError(self)
ConnectClosedError: ConnectClosedError(ip-172-31-9-153.us-east-2.compute.internal)
[INFO    ] Raising ConnectionClosedException instead
[ERROR   ] Traceback (most recent call last):
  File "/home/admin/salt/salt/utils/napalm.py", line 153, in call
    out = getattr(napalm_device.get('DRIVER'), method)(*args, **kwargs)
  File "/home/admin/napalm-base/napalm_base/base.py", line 59, in fun
    raise err_class(str(error))
ConnectionClosedException: ConnectClosedError(ip-172-31-9-153.us-east-2.compute.internal)
```


napalm_base/base.py
```diff
@@ -16,20 +16,71 @@
...
 import napalm_base.exceptions
 import napalm_base.helpers
-
 import napalm_base.constants as c
-
 from napalm_base import validate
 
+log = logging.getLogger(__file__)
+
+ERROR_MAP = {
```
dbarrosop: This should be a class attribute and it should be empty for the base driver. Then each driver can implement it's own ERROR_MAP.

napalm_base/base.py
```diff
@@ -16,20 +16,71 @@
...
+        except Exception as error:
+            err_name = error.__class__.__name__
+            err_mod = error.__class__.__module__
+            err_full_name = '{}.{}'.format(err_mod, err_name)
+            log.error('Raised {}'.format(err_full_name), exc_info=True)
+            if err_full_name in ERROR_MAP:
+                err_class = ERROR_MAP[err_full_name]
+                log.info('Raising {} instead'.format(err_class.__name__))
+                raise err_class(str(error))
+            # Raise everything else, using the original class.
```
dbarrosop: It would be great if the `else` statement here would do something like this:

https://github.com/napalm-automation/napalm-base/blob/develop/napalm_base/base.py#L87

dbarrosop: Potentially raising an `UncaughtException` or something like that. Something we can distinguish easily.

napalm_base/base.py
mirceaulinic: Yes, makes sense. Just that I'm not sure yet what's the best approach... this might be a chicken and egg type thing (napalm-base depends on the driver, which depends on napalm-base). I will investigate.

napalm_base/base.py
mirceaulinic: Sounds good.

napalm_base/base.py
dbarrosop: > napalm-base depends on the driver

But it doesn't. It's just a class method that uses a class attribute. Your subclass just happens to override the class attribute. The parent class has no idea about it :)

mirceaulinic:
@dbarrosop @ktbyers pushed the changes as suggested.

The current approach requires the driver to define a map like:

```python
ERROR_MAP = {
        'jnpr.junos.exception.ConnectClosedError': napalm_base.exceptions.ConnectionClosedException,
        'jnpr.junos.exception.ConnectAuthError': napalm_base.exceptions.ConnectAuthError,
        'jnpr.junos.exception.ConnectTimeoutError': napalm_base.exceptions.ConnectTimeoutError,
        'jnpr.junos.exception.LockError': napalm_base.exceptions.LockError,
        'jnpr.junos.exception.UnlockError': napalm_base.exceptions.UnlockError,
        'jnpr.junos.exception.CommitError': napalm_base.exceptions.CommitError
    }
```

We can have either have the mapping between the full path, either 1:1 class, either both. What do you think?

dbarrosop:
I would import the classes directly to avoid issues that are only triggered at runtime because someone misspelled a path or class name.

mirceaulinic:
I'd fine with a decorator.

FYI: this is exactly what this metaclass does - decorates all the methods (with `_raise_napalm_error`), without requiring the user to decorate them manually.
(I'm not keen on implementing it like that, and I agree it's less obvious)

ktbyers:
Yes, understood...it is a trade-off of slightly more work versus hidden-behind-the-scenes.

mirceaulinic:
I'm fine with any of the approaches. So @ktbyers @dbarrosop please let me know which one would be preferable (if you see any advantages/disadvantages).
No rush on this, we can leave it aside a few days and think more about it...

ktbyers:
Sounds good...

dbarrosop:
I'd rather use the metaclass to avoid "forgetting" about it. Just make sure that the metaclass and the wrapper function are well documented and it should be fine. The traceback should show the wrapper function if something goes bad so not super concerned about the whole stuff failing without being super explicit about it.

ktbyers:
I won't put up any more objections on this decorator vs metaclass...

dbarrosop:
@mirceaulinic when you have the time link a PR using this. Would be great to have them hand in hand to see how it's used :)

bewing:
Looks like https://pythonhosted.org/six/#six.reraise is needed for full compatibility between py2 and py3 for re-raising exceptions.

Standalone code in https://stackoverflow.com/questions/34463087/raise-exception-in-python-2-x-and-3-x

mirceaulinic:
Hi @bewing - thanks for suggestion. I recall I had some changes not pushed yet, hope I didn't stash them. Let me revisit this somewhere next week.

PR #261, opened 2017-06-06
Documentation - Fix mac address format to be proper form.

ktbyers:


napalm_base/base.py
```diff
@@ -247,6 +247,7 @@ def get_interfaces(self):
                 'last_flapped': -1,
                 'speed': 1000,
                 'mac_address': u'dead:beef:dead',
```
bewing: This line wasn't removed?

ktbyers: ? I just checked in the develop branch and in the PR, that was removed?

It still needs updated in master branch as those changes haven't been incorporated in the master branch.

bewing: Sorry, was working through notifications, and didn't notice it took me to a single commit, and not the whole changeset from the PR

PR #263, opened 2017-06-08
Improve coverage of get_network_driver()

bewing:
Not sure if using napalm-logs to test a real module not having NetworkDriver is correct, or if we should create a specific dummy module to test.  Also test a non-string type to hit 86, napalm_eos to hit cover 94->96 jump.

### DIFF
--- a/napalm_base/__init__.py
+++ b/napalm_base/__init__.py
@@ -49,7 +49,7 @@ __all__ = [
 ]
 
 
-def get_network_driver(module_name):
+def get_network_driver(module_name, prepend=True):
     """
     Searches for a class derived form the base NAPALM class NetworkDriver in a specific library.
     The library name must repect the following pattern: napalm_[DEVICE_OS].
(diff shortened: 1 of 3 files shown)


bewing:
Looks like a specific package would be better, as napalm-logs isn't Python3 ready. :(

mirceaulinic:
@bewing please remove napalm-logs from the requirements - this is not a regular driver: see the readme (although there's no proper documentation): https://github.com/napalm-automation/napalm-logs

bewing:
@mirceaulinic The point was to have a napalm_ module that does NOT contain a NetworkDriver() method, to improve test coverage in __init__.py.  If it's just navel-gazing, I'm OK with rejecting the PR.

dbarrosop:
@bewing what branch of the code you want to test?

bewing:
@dbarrosop https://github.com/napalm-automation/napalm-base/blob/develop/napalm_base/__init__.py#L109

dbarrosop:
Ok, you can probably use any module already imported. Like the `sys` module. It doesn't necessarily have to belong to napalm.

bewing:
Tried that already - if the module doesn't start with napalm_, it's prepended, so passing "sys" means we hit the non-existent module check instead of the "real module, no driver" check

dbarrosop:
> if the module doesn't start with napalm_

We should probably fix that. Maybe we can add a kw argument to avoid prepending?

```
get_network_driver(name, prepend=True)
```

bewing:
Sorry for the delay -- had some spam folder issues.

dbarrosop:
LGTM @ktbyers @mirceaulinic any comments?

PR #264, opened 2017-06-10
Added integration with napalm-yang

dbarrosop:
DO NOT MERGE

Depends on napalm-automation/napalm-yang#61

This PR integrates napalm-yang into the drivers. See:

https://github.com/napalm-automation/napalm-base/compare/dbarrosop/napalm-yang?expand=1#diff-b284a28710cce90d9d9be3a7f4cabc8e

(gist with the execution https://gist.github.com/dbarrosop/8fae309ecbfe7d1207044aca5732beba)

"getters" are dynamic which means there is no dependency loop between napalm-yang and the drivers themselves. I just have to move the `SUPPORTED_MODELS` to `napalm-yang` which I will do once we are happy with the PR.

@mirceaulinic @ktbyers thoughts?

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -22,13 +22,14 @@ import sys
 # local modules
 import napalm_base.exceptions
 import napalm_base.helpers
+import napalm_base.yang
 
 import napalm_base.constants as c
 
 from napalm_base import validate
 
 
-class NetworkDriver(object):
+class NetworkDriver(napalm_base.yang.NapalmYangIntegration):
 
     def __init__(self, hostname, username, password, timeout=60, optional_args=None):
         """
--- a/napalm_base/yang.py
+++ b/napalm_base/yang.py
@@ -1,0 +1,105 @@
+# Copyright 2017 Dravetech AB. All rights reserved.
+#
+# The contents of this file are licensed under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with the
+# License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
+# Python3 support
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import napalm_yang
+
+
+# TODO this should come from napalm-yang
+# TODO we probably need to adapt the validate framework as well
+SUPPORTED_MODELS = [
+    "openconfig-interfaces",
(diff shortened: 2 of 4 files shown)


dbarrosop:
Pinging @matejv and @bewing as I know they have been playing with `napalm-yang` as well.

napalm_base/base.py
```diff
@@ -93,6 +94,13 @@ def __raise_clean_exception(exc_type, exc_value, exc_traceback):
         # Traceback should already be attached to exception; no need to re-attach
         raise exc_value
 
+    @property
+    def yang(self):
+        if not hasattr(self, "config"):
+            self.config = napalm_base.yang.Yang("config", self)
+            self.state = napalm_base.yang.Yang("state", self)
+        return self
+
```
dbarrosop: Yes, so that's certainly a hack. Ideally we would have something like:

```
class Foo(object):
    def __init__(self):
        self.config = napalm_base.yang.Yang("config", self)
        self.state = napalm_base.yang.Yang("state", self)

class NetworkDriver(object):

        def __init__(self):
             self.yang = Foo()
```

But right now we can't do that because `__init__` throws a `NotImplementedException` and no driver calls super for that reason. So to make that happen we would have to change all drivers. With this hack I managed to provide same behavior without having to change any driver.

If you have a better solution let me know, this is the best I could come up with so far. We could probably use a wrapper or a metaclass but figured this was simpler/easier to read (if we go with this hack I will add details to the docstrings so people reading this knows why the hack).

dbarrosop:
I have added a mechanism to print the models as well, I guess it's useful for reference.

napalm_base/base.py
```diff
@@ -22,6 +22,7 @@
 # local modules
 import napalm_base.exceptions
 import napalm_base.helpers
+import napalm_base.yang
```
mirceaulinic: If we want to avoid `napalm-yang` as a dependency for `napalm-base`, we can check if it is installed, i.e.:

```python
try:
    import napalm_yang
    HAS_NY = True
except ImportError:
    HAS_NY = False
```

napalm_base/base.py
```diff
@@ -93,6 +94,13 @@ def __raise_clean_exception(exc_type, exc_value, exc_traceback):
         # Traceback should already be attached to exception; no need to re-attach
         raise exc_value
 
+    @property
+    def yang(self):
+        if not hasattr(self, "config"):
```
mirceaulinic: Just before this, we can:

```python
if not HAS_NY:
    raise SomeException('Please install napalm-yang for this fancy stuff')
```

mirceaulinic:
Okay, Python 2 tests passed here. I think this only waits for python 3 support in napalm-yang @dbarrosop?

dbarrosop:
I won't wait until napalm-yang has support for py3 because that will just delay things and `napalm-yang` is an optional module anyway. However, as I write some documentation on how to use this I am making minor adjustments so hold on this for a few more days.

PR #266, opened 2017-06-20
Correctly check all keys comply

dbarrosop:


bewing:
Just checking the last one's not good enough?

Issue #267, opened 2017-06-30
is_alive() function in base.py is throwing an AttributeError even before open() is called


ipv6freely:
Im getting an AttributeError exception in the is_alive() function that is used for garbage collection in napalm-base 0.24.1.

**Code:**
```
#!.venv/bin/python
import napalm
"""
napalm (1.2.0)
napalm-base (0.24.1)
napalm-junos (0.10.3)
"""

def napalm_this_bitch(password_list, name, ipaddress, napalm_driver):
    for password in password_list:
        print(f'{name:40} {ipaddress:20} {napalm_driver:10} {password:10}')
        driver = napalm.get_network_driver(napalm_driver)
        device = driver(hostname=ipaddress, username='admin', password=password, optional_args={'port': 22})

password_list = ['foo']
device_list = [{'name':'TESTDEVICE-1','ipaddress':'10.212.72.7','napalm_driver':'junos'}]
for device in device_list:
    napalm_this_bitch(password_list, device['name'], device['ipaddress'], device['napalm_driver'])
```

**Output:**
```
cjones@cjones-mbp$ python test.py
TESTDEVICE-1                             10.212.72.7          junos      foo
Exception ignored in: <bound method NetworkDriver.__del__ of <napalm_junos.junos.JunOSDriver object at 0x1021b9f98>>
Traceback (most recent call last):
  File "/Users/cjones/python/pwchanger/.venv/lib/python3.6/site-packages/napalm_base/base.py", line 69, in __del__
    if self.is_alive()["is_alive"]:
  File "/Users/cjones/python/pwchanger/.venv/lib/python3.6/site-packages/napalm_junos/junos.py", line 156, in is_alive
    'is_alive': self.device._conn._session.transport.is_active() and self.device.connected
AttributeError: 'NoneType' object has no attribute '_session'
```

ipv6freely:
I just submitted a pull request with an easy fix.

ipv6freely:
Re-submitted with an exception handler catching all.

PR #268, opened 2017-06-30
Added AttributeError Exception to __del__ function in NetworkDriver

ipv6freely:
This commit fixes Issue #267:
[https://github.com/napalm-automation/napalm-base/issues/267](url)

Adding an exception handler for AttributeError allows the script to complete.

### DIFF
--- a/napalm_base/base.py
+++ b/napalm_base/base.py
@@ -68,6 +68,8 @@ class NetworkDriver(object):
         try:
             if self.is_alive()["is_alive"]:
                 self.close()
+        except AttributeError:
+            pass
         except NotImplementedError:
             pass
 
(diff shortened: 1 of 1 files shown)


napalm_base/base.py
```diff
@@ -68,6 +68,8 @@ def __del__(self):
         try:
             if self.is_alive()["is_alive"]:
                 self.close()
+        except AttributeError:
```
dbarrosop: As this is triggered only when deleting the object I think it's actually safer to just capture all the exception. Would you mind replacing both `except clauses with just one`?

```
         try:
              if self.is_alive()["is_alive"]:
                  self.close()
         except Exception:
              pass
```

ipv6freely:
Yes I agree with you. I have modified that code.

dbarrosop:
Thanks!

dbarrosop:
Fixes automation/napalm-base/issues/267

PR #279, opened 2017-07-16
Tests for config under snmp-server (for #277)

ogenstad:
An additional test for #277. It's nice that the code knows how to handle the added `snmp-server community public RO 10` line correctly.

However it doesn't work for the added `snmp-server enable traps entity-sensor` as it thinks it's supposed to delete `snmp-server enable traps entity`. 

There will be a number of these scenarios, probably "a lot". In some instances it will also surely differ between various platforms.

### DIFF
--- a/napalm_base/indent_differ.py
+++ b/napalm_base/indent_differ.py
@@ -66,6 +66,7 @@ def _can_have_multiple(command):
         "neighbor",
         "ip address",
         "ipv6 address",
+        "snmp-server enable traps",
     ]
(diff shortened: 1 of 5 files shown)


ogenstad:
Adding `snmp-server enable traps`, to the EXACT_MATCHES list in `_can_have_multiple()` made it work as expected. Should that list be a common list for all platforms? Perhaps it would be cleaner to move the list to its own file as it grows.

PR #280, opened 2017-07-16
Test multiple identical remarks (for #277)

ogenstad:
Another test for #277 which uses multiple identical remarks. Don't know how common this would be, only thinking about scenarios where the differ might be struggling.

### DIFF
--- a/napalm_base/indent_differ.py
+++ b/napalm_base/indent_differ.py
@@ -69,6 +69,7 @@ def _can_have_multiple(command):
         "ip address",
         "ipv6 address",
         "snmp-server enable traps",
+        "remark",
     ]
     return any([command.startswith(e) for e in EXACT_MATCHES])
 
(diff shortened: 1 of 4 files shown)


dbarrosop:
Will be merged via #277

dbarrosop:
Oh, I see what you did here. Goddamn you, you broke it!!!

I will think about it but I'd say it's fine if the diff shows only:

```
  ip access-list extended TEST-ACL
+   remark # DENY
+   deny  ip any any
```

As it's the case. The solution might be too ugly for little gain as this is mostly aesthetics, the unreported command actually doesn't give you any useful information and I am sure nobody would miss it :P

Anyway, will scratch my head around this for a while...

ogenstad:
Mission accomplished ;)

But yeah, I agree with you. Don't know if there's a real problem here, just a strange thing which might pop up sometime. Can't think of any other scenario which would look like this one.

dbarrosop:
Closing this one for the time being.

PR #282, opened 2017-07-16
Load tests automatically

ogenstad:
@dbarrosop, I'm guessing there might be a reason why you didn't do this from the start. Anyway it feels like it should be there.


dbarrosop:
Yes, laziness :)

Thx!

PR #284, opened 2017-07-18
Convert napalm_base testing to tox

bewing:
Convert napalm_base to tox testing.

py27 tests clean, I don't have py34 installed locally, and the new mock driver fails tests on 3.5 and 3.6.

Will dig into tomorrow.


bewing:
@ktbyers @dbarrosop 
If you have the spare time to investigate the py3 errors in test_mock_driver, I'd appreciate it, otherwise I'll try to get to it this week/weekend

ogenstad:
Try changing this in `test_basic()` in test/unit/test_mock_driver.py

From:
```python
        with pytest.raises(napalm_base.exceptions.ConnectionClosedException) as excinfo:
            d.get_facts()
        assert "connection closed" in excinfo.value
```

To:
```python
        with pytest.raises(napalm_base.exceptions.ConnectionClosedException) as excinfo:
            d.get_facts()
        assert "connection closed" in str(excinfo.value)
```

Looks like it could be the same for the other errors.

requirements-dev.txt
```diff
@@ -1,8 +1,20 @@
...
 pytest
 pytest-cov
 pytest-json
 pytest-pythonpath
 pylama
 flake8-import-order
-nose
+ddt
+lxml
+napalm-eos
```
dbarrosop: why are we adding the drivers here?

test/unit/conftest.py
```diff
@@ -0,0 +1,40 @@
+import pytest
```
dbarrosop: What's the purpose of this file here? At first I thought this was for the MockDriver but I don't think it is :P

bewing: I'm actually kind of confused about the split between requirements-dev.txt (which should have all packages required to run tests / development) and the test/unit/requirements.txt file.  Should they be combined?   Should we have the expectation that just installing requirements-dev.txt and running py.test should have failed TestGetNetworkDriver() tests?

I had moved all the libraries up from test/unit/requirements.txt to requirements-dev.txt, but would have no issues with moving them back.

requirements-dev.txt
bewing: I'm actually kind of confused about the split between requirements-dev.txt (which should have all packages required to run tests / development) and the test/unit/requirements.txt file. Should they be combined? Should we have the expectation that just installing requirements-dev.txt and running py.test should have failed TestGetNetworkDriver() tests?

I had moved all the libraries up from test/unit/requirements.txt to requirements-dev.txt, but would have no issues with moving them back.

test/unit/conftest.py
bewing: I... honestly don't remember.  I was having some issues, and thought this solved it, but pulling the file completely doesn't change the test results.  It might have been a legacy of me running tests on my napalm-base wrapper package (where I stole most of this code from)

dbarrosop: Let's remove it then. I don't think it's needed here.

requirements-dev.txt
dbarrosop: Well, the reasoning is you don't really need those libs for development purposes. I, for example, prefer to install with `pip install -e .` the individual drivers as then I have an egg linked to the local repo so I can test changes as I type them. But a personal preference shouldn't dictate how things are done in the project.

In any case, this is a moot discussion though as we will be merging the drivers soon. So we can leave this here.

bewing:
Should we write tests for napalm_base/tests/*, or just omit them from the coverage report?

ktbyers:
@bewing I think we should omit them from coverage report.

test/unit/test_mock_driver.py
```diff
@@ -39,7 +39,7 @@ def test_basic(self):
 
         with pytest.raises(napalm_base.exceptions.ConnectionClosedException) as excinfo:
             d.get_facts()
-        assert "connection closed" in excinfo.value
+        assert "connection closed" in str(excinfo.value)
```
dbarrosop: I guess we should be using here `py23_compat.text_type` instead

dbarrosop:
It's nice to see the last bump :P

dbarrosop:
LGTM, @ktbyers any objections?

dbarrosop:
Let's merge it! :D

PR #288, opened 2017-08-01
Fix napalm validate so identical strings will work

ktbyers:
Essence of change is this:

```
+            m = re.search(src, py23_compat.text_type(dst))
+            if m:
+                return bool(m)
+            else:
+                return src == dst
```

Basically try re.search first and if that fails, then look for exact string comparison.

I also eliminated an unnecessary 'else' from the code as it was serving no purpose.


dbarrosop:
LGTM, merging

PR #289, opened 2017-08-01
Correct link under "compliance_report" function

bdlamprecht:
Fixing a bad link in the documentation for "compliance_report" function.


dbarrosop:
Thanks a lot!

dbarrosop:
Note you won't see the fix until we don't release a new `napalm-base` version.

PR #293, opened 2017-08-04
DO NOT MERGE - Search MRO path for templates

bewing:
@mirceaulinic Does this break anything in salt?


mirceaulinic:
As long as you can still do:

```
d = get_network_driver('X')
i = d(host, usr, pass)
i.load_template(...)
```

Nothing is affected.

bewing:
There's a logic bug in here -- the helper uses the class passed to determine what load_configuration method to call.  Need to rework this branch to just expand the searchpath inside the helper function

PR #295, opened 2017-08-05
Mock return text

dbarrosop:


dbarrosop:
Do not merge. If we merge #296 I will probably scrap the Mockdriver and replace it with the `Recorder`.

PR #308, opened 2017-09-11
Add tests for validation from source

ubaumann:
Update base.py and add tests.
Remember the change is backwards incompatible (default value has changed)


dbarrosop:
LGTM however, there are a few minor pylama issues, could you fix them, please?

ubaumann:
I'm so sorry. I should not open a PR after midnight.

PR #311, opened 2017-09-20
What about a get_ipv6_neighbors_table() method?

pierky:


mirceaulinic:
Hi @pierky - thanks for this PR.
We actually do have an implementation available for this this method: https://github.com/napalm-automation/napalm-eos/pull/164, but we didn't have the test definition and the header.
Your proposal looks good to me, as it is in-line with the `get_arp_table` brother.
Please fix the conflict and then I think we are good to go.
@napalm-automation/council any opinions?

pierky:
Hi @mirceaulinic, I didn't notice the EOS implementation, sorry.

The output format I proposed here is different from the one used in napalm-automation/napalm-eos#164: how do we proceed now?

mirceaulinic:
You are correct @pierky , the EOS implementation is not the same. But I like yours as it is very similar to the `get_arp_table` structure, so I would suggest to move forward with yours.

Note that the structure however it is not complete (due to a design fault we made in `get_arp_table`), but I vote for this to be accepted so we have the IPv6 equivalent for the `get_arp_able`. The complete version will probably be covered using the yang models.

ktbyers:
I assume age is in seconds?

Besides this it looks good to me.

pierky:
@ktbyers I explicitly stated it within the docstring now. And also fixed the IOS implementation ;-)
Thanks for pointing that out.

dbarrosop:
LGTM, I also prefer this one as it's closer to the ipv4 equivalent.

PR #312, opened 2017-09-24
Fixing issue with PY2 'long' failure in unit test

ktbyers:


ktbyers:
With the old code if `data[key]` returned a type long, then correct_class would get set to False here

```
correct_class = isinstance(data[key], instance_class) and correct_class
```

And would never be fixed...so unit test would fail on a 32-bit machine.

PR #319, opened 2017-11-07
Migrate napalm base

mirceaulinic:


ktbyers:
I am pretty sure this is different than what we agreed to over the summer (though not entirely sure of this). 

I thought what we agreed to was only to previously was only modify get_network_driver() so that it would load napalm2 first and then load legacy napalm drivers?

These changes will potentially break people's existing code when they `pip install napalm`. 

Basically, you are trying to bring the new napalm into the old napalm base and I don't think we should try to do this. This is a pattern that has the potential for creating a lot of problems and confusion.

Excluding get_network_driver() napalm_base should be frozen in time. If people want to use the new napalm, they will have to upgrade.

Can you explain why we would want to do this? 

What are some examples where using imports from napalm_base with the new napalm are required (excluding get_network_driver)?

mirceaulinic:
@ktbyers @dbarrosop I have revisited this and we'll be importing `get_network_driver` from napalm>=2.0. Please let me know if you have any further suggestions / comments.

ktbyers:
@mirceaulinic So are you just going to close this PR and re-submit it?

mirceaulinic:
Hi @ktbyers - I have pushed the changes as discussed. in this branch. Let me know if there's anything I missed.

napalm_base/__init__.py
```diff
@@ -20,10 +20,11 @@
 
 # Python std lib
 import sys
-import inspect
-import importlib
 import pkg_resources
 
+# Import local modules
+from napalm_base.base import NetworkDriver
```
ktbyers: I would think we should have `get_network_driver` and `NetworkDriver` coupled together i.e. we should import both from napalm2 or import both old napalm_base.

napalm_base/__init__.py
```diff
@@ -32,83 +33,104 @@
...
+            # Can also request using napalm_[SOMETHING]
+            if 'napalm_' not in module_install_name and prepend is True:
+                module_install_name = 'napalm_{name}'.format(name=module_install_name)
+            module = importlib.import_module(module_install_name)
+        except ImportError:
+            raise ModuleImportError(
+                    'Cannot import "{install_name}". Is the library installed?'.format(
+                        install_name=module_install_name
+                    )
+                )
```
ktbyers: I assume all of this get_network_driver code is the same as what we had previously just indented under the `else`?

requirements.txt
```diff
@@ -1,4 +1 @@
-jtextfsm
-jinja2
-netaddr
-pyYAML
+napalm>=2.0.0
```
ktbyers: Do we want to do that? I would think we wouldn't since this would force upgrade them to the new napalm. In other words, this would make them go to the reunified napalm when they might not be aware of it.

I think it would be better to print a warning message?

napalm_base/__init__.py
mirceaulinic: I agree.

napalm_base/__init__.py
mirceaulinic: I agree.

napalm_base/__init__.py
mirceaulinic: Yes, nothing changed there, just the entire block indented.

requirements.txt
mirceaulinic: Makes sense.

mirceaulinic:
@ktbyers I updated this branch, can you please re-review when you got a minute?

ktbyers:
Looks good +1

